PR #8, opened 2015-06-30
[SYNCOPE-675] Role tree builder now performs a lazy load of roles

giacomolm:
See https://issues.apache.org/jira/browse/SYNCOPE-675 for details. 
It is worth noting I'm exposing children service, with few modification. In the current implementation of this method (in Syncope), owned entitlements are controlled: in this pull request I removed it. In fact the list method - that in my opinion has a larger functionality - doesn't include this check.


giacomolm:
[SYNCOPE-676] made this issue outdated

PR #15, opened 2016-05-04
[SYNCOPE-799] user Admin cannot log in to enduser anymore

mteo:


ilgrosso:
LGTM! I'll go ahead an merge.

PR #17, opened 2016-05-13
[SYNCOPE-783] date and time fields management added in Syncope-enduser

mteo:


client/enduser/src/main/resources/META-INF/resources/app/js/directives/dynamicPlainAttribute.js
```diff
@@ -16,6 +16,8 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+/* global moment, date */
```
ilgrosso: Please remove this comment, since it seems that `moment` is not used any more.

client/enduser/src/main/resources/META-INF/resources/app/js/directives/dynamicPlainAttribute.js
```diff
@@ -168,12 +171,27 @@ angular.module('self')
...
+                    $scope.ismeridian = true;
+                    $scope.toggleMode = function () {
+                      $scope.ismeridian = !$scope.ismeridian;
+                    };
                     break;
 
                   case "Boolean":
                     $scope.user.plainAttrs[schema.key].values[index] =
-                           $scope.user.plainAttrs[schema.key].values[index] == "true" ? true : false;
+                            $scope.user.plainAttrs[schema.key].values[index] == "true" ? true : false;
```
ilgrosso: Is this really needed this way? Can't you change it to (at least):

```javascript
$scope.user.plainAttrs[schema.key].values[index] = 
  $scope.user.plainAttrs[schema.key].values[index] == "true";
```

ilgrosso:
Since you are not using `momentjs` anymore - as it seems - please remove any dependency from root `pom.xml` and `client/console/pom.xml`.

The rest looks good, thanks.

ilgrosso:
LGTM! Merged.

PR #18, opened 2016-06-03
Added Exception for existing key (SYNCOPE-866)

tmess567:
I have added a new Exception named **KeyAlreadyExistsException** and attached it to the **MailTemplateLogic** and **ReportTemplateLogic** classes to check for duplicate keys and throw the new Exception.

I have also tested this out on Swagger and attached screenshots of the error for both **MailTemplates**

mailtemplatekeyalreadyexistsexception

and **ReportTemplates**.

reporttemplatekeyalreadyexistsexception

### DIFF
--- a/core/logic/src/main/java/org/apache/syncope/core/logic/MailTemplateLogic.java
+++ b/core/logic/src/main/java/org/apache/syncope/core/logic/MailTemplateLogic.java
@@ -31,6 +31,7 @@ import org.apache.syncope.common.lib.types.ClientExceptionType;
 import org.apache.syncope.common.lib.types.MailTemplateFormat;
 import org.apache.syncope.common.lib.types.StandardEntitlement;
 import org.apache.syncope.core.persistence.api.dao.NotFoundException;
+import org.apache.syncope.core.persistence.api.dao.DuplicateException;
 import org.apache.syncope.core.persistence.api.dao.MailTemplateDAO;
 import org.apache.syncope.core.persistence.api.dao.NotificationDAO;
 import org.apache.syncope.core.persistence.api.entity.EntityFactory;
@@ -83,6 +84,9 @@ public class MailTemplateLogic extends AbstractTransactionalLogic<MailTemplateTO
 
     @PreAuthorize("hasRole('" + StandardEntitlement.MAIL_TEMPLATE_CREATE + "')")
     public MailTemplateTO create(final String key) {
+        if (mailTemplateDAO.find(key) != null) {
+            throw new DuplicateException(key);
+        }
         MailTemplate mailTemplate = entityFactory.newEntity(MailTemplate.class);
         mailTemplate.setKey(key);
         mailTemplateDAO.save(mailTemplate);
--- a/core/logic/src/main/java/org/apache/syncope/core/logic/ReportTemplateLogic.java
+++ b/core/logic/src/main/java/org/apache/syncope/core/logic/ReportTemplateLogic.java
@@ -31,6 +31,7 @@ import org.apache.syncope.common.lib.types.ClientExceptionType;
 import org.apache.syncope.common.lib.types.ReportTemplateFormat;
 import org.apache.syncope.common.lib.types.StandardEntitlement;
 import org.apache.syncope.core.persistence.api.dao.NotFoundException;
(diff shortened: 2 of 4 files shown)


core/logic/src/main/java/org/apache/syncope/core/logic/MailTemplateLogic.java
```diff
@@ -83,6 +84,13 @@ public MailTemplateTO transform(final MailTemplate input) {
 
     @PreAuthorize("hasRole('" + StandardEntitlement.MAIL_TEMPLATE_CREATE + "')")
     public MailTemplateTO create(final String key) {
+        for (MailTemplateTO mailTemplateTO: list()) {
```
ilgrosso: Take a look at [ResourceLogic](https://github.com/apache/syncope/blob/master/core/logic/src/main/java/org/apache/syncope/core/logic/ResourceLogic.java#L121-L123) and do accodingly.

core/logic/src/main/java/org/apache/syncope/core/logic/ReportTemplateLogic.java
```diff
@@ -84,6 +85,12 @@ public ReportTemplateTO transform(final ReportTemplate input) {
 
     @PreAuthorize("hasRole('" + StandardEntitlement.REPORT_TEMPLATE_CREATE + "')")
     public ReportTemplateTO create(final String key) {
+        for (ReportTemplateTO reportTemplateTO: list()) {
```
ilgrosso: Take a look at [ResourceLogic](https://github.com/apache/syncope/blob/master/core/logic/src/main/java/org/apache/syncope/core/logic/ResourceLogic.java#L121-L123) and do accodingly.

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/KeyAlreadyExistsException.java
```diff
@@ -0,0 +1,35 @@
+/*
```
ilgrosso: This class is not needed, please remove and use instead `org.apache.syncope.core.persistence.api.dao.DuplicateException` (see above).

ilgrosso:
Thanks for this PR @tmess567, I appreciate it very much.
I have added some comments to avoid duplicate classes and stay close to what other Logic classes do.

Please fix with an additional commit.
Thank you very much.

ilgrosso:
Please also add a test case on [ReportTemplateITCase](https://github.com/apache/syncope/blob/master/fit/core-reference/src/test/java/org/apache/syncope/fit/core/ReportTemplateITCase.java) and on [MailTemplateITCase](https://github.com/apache/syncope/blob/master/fit/core-reference/src/test/java/org/apache/syncope/fit/core/MailTemplateITCase.java) named `issueSYNCOPE866()`.

tmess567:
So, in the crud method, I will be calling the issueSYNCOPE866() method? Or should I just add a Test Case directly into the crud method?

ilgrosso:
Ok, let's do things one step at a time.
Please first address my comments in code, we will come up with testing later on, ok?

tmess567:
That's done. I'll commit right away.

ilgrosso:
Ok, the changes are just fine.

Now, you need to add a couple of methods like as follows

```java
@Test
public void issueSYNCOPE866() {

}
```

to the two classes mentioned above.

In the body of such methods you will, by taking reference with other methods in the same class, essentially:

1. create a report template with an existing key
1. assert that such creation operation fails with

The following is a working sample - that you can just grab and put there - for report template:

```java
    @Test
    public void issueSYNCOPE866() {
        ReportTemplateTO reportTemplateTO = new ReportTemplateTO();
        reportTemplateTO.setKey("empty");
        try {
            reportTemplateService.create(reportTemplateTO);
            fail();
        } catch (SyncopeClientException e) {
            assertEquals(ClientExceptionType.EntityExists, e.getType());
        }
    }
```

tmess567:
Done. Please check.

ilgrosso:
LGTM, please squash your three commits into one, and don't forget to mention SYNCOPE-866 in the final commit log (the very first used here will work, anyway).

I will merge this PR at earliest next Monday 6th, after that the freeze period due to the ongoing release VOTE is over.

PR #34, opened 2016-08-17
Added Tests for Eclipse Plugin

tmess567:
I have yet to test out the Travis config to see if it can run these tests.
Also Editor Tests are not complete yet because for some reason swtbot is unable to pick up the resource bundle for the HTMLEditor.

These tests check if the admin login works and whether or not mail and report templates can be added and deleted using the tree view.

### DIFF
--- a/ide/eclipse/bundles/org.apache.syncope.ide.eclipse.tests/pom.xml
+++ b/ide/eclipse/bundles/org.apache.syncope.ide.eclipse.tests/pom.xml
@@ -1,0 +1,154 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.syncope.ide</groupId>
+    <artifactId>syncope-ide-eclipse</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+    <relativePath>../../</relativePath>
+  </parent>
+
+  <name>Apache Syncope IDE Eclipse Tests</name>
+  <description>Apache Syncope IDE Eclipse Tests</description>
+  <groupId>org.apache.syncope.ide.eclipse</groupId>
+  <artifactId>org.apache.syncope.ide.eclipse.tests</artifactId>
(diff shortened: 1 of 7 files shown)


ilgrosso:
Linking to SYNCOPE-809

ide/eclipse/bundles/org.apache.syncope.ide.eclipse.tests/pom.xml
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ilgrosso: Why tests in a separate Maven module?

ide/eclipse/bundles/org.apache.syncope.ide.eclipse.tests/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+        <configuration>
+          <useUIHarness>true</useUIHarness>
+          <useUIThread>false</useUIThread>
+          <product>org.eclipse.sdk.ide</product>
+          <application>org.eclipse.ui.ide.workbench</application>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
```
ilgrosso: Where is the `skipTests` profile?

ide/eclipse/bundles/org.apache.syncope.ide.eclipse.tests/src/main/java/org/apache/syncope/ide/eclipse/tests/SyncopeViewTestCase.java
```diff
@@ -0,0 +1,338 @@
...
+    private SWTBotTreeItem mTemplateItem;
+    private SWTBotTreeItem rTemplateItem;
+    private SWTBotTreeItem[] mtemplateList;
+    private SWTBotTreeItem[] rtemplateList;
+
+    protected static final String ADMIN_UNAME = "admin";
+
+    protected static final String ADMIN_PWD = "password";
+
+    protected static final String ADDRESS = "http://localhost:9080/syncope/rest";
```
ilgrosso: Do such test cases require Apache Syncope core running? (I think so)
If so, you should refactor things to include your test cases under `fit/core-reference/src/test/java`

ilgrosso:
Overall, as you can read from comments, I think it would be better to move the test cases under `fit/core-reference`.
Moreover, isn't there any way to run such tests headless, e.g. without requiring Xvfb and `DISPLAY`?

ide/eclipse/bundles/org.apache.syncope.ide.eclipse.tests/pom.xml
tmess567: Because to run the tests, surefire needs the plugin module to be already built.

tmess567:
Most of the project is mainly integration with eclipse UI so headless tests wouldn't help much.

tmess567:
The test cases also require eclipse plugin to be built and so should be kept separate from `fit/core-reference` and after `Apache Syncope IDE Eclipse Plugin` module.

ilgrosso:
I'd suggest then to move tests to a dedicated profile in fit/core-reference (as running core is requested), with test dependency of the eclipse plugin.

Tomorrow I can provide more details, see you on IRC

ilgrosso:
You can find [attached](https://github.com/apache/syncope/files/424433/updated34.diff.zip) a diff file I have generated starting from your modifications here.

Basically, I have added the tests module and execution in a dedicated `eclipse-it` profile, **not** active by default: this means that such tests are not ordinarily executed. I have reverted your changes to `travis.yml` for this reason.

Now, after running `mvn -PskipTests,all` from the root directory, you can move to `ide/eclipse/bundles/org.apache.syncope.ide.eclipse.tests` and launch `mvn -P eclipse-it`: this will first start the Syncope core locally via cargo, then execute the available tests.

I have tried on my laptop and it works.

You should now revert your modifications here and manually apply the attached diff file and check if everything is fine.

Thanks!

ilgrosso:
There was some problem with yesterday's patch: 
[this new one](https://github.com/apache/syncope/files/426824/new34.diff.zip) can be applied on top of your existing repo.

ilgrosso:
Just merged, thank you

PR #35, opened 2016-08-22
Fixing HTMLEditor Resource Bundle location SYNCOPE-809

tmess567:
This also fixes the testing problem with the HTMLEditor as stated in #34 .
However, the StructuredTextEditor still malfunctions during the swtbot tests.


tmess567:
Closed with 5c304ad72dcd6d71a97cc73f9f3e96b1aac39d9b

PR #36, opened 2016-08-29
Startup performance tweaks : restrict startup scanning to syncope packages

gonzalad:
Without this PR, standalone tomcat startup takes 135s (by default, syncope scans all packages).
With this PR, it takes 32s.

Startup scanning is restricted to packages : 
 * org.apache.syncope.client.console
 * org.apache.syncope.fit.core
 * org.apache.syncope.core

### DIFF
--- a/client/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
+++ b/client/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
@@ -20,12 +20,14 @@ package org.apache.syncope.client.console.init;
 
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import org.apache.commons.collections4.ComparatorUtils;
 import org.apache.commons.lang3.ArrayUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.syncope.client.console.pages.BaseExtPage;
 import org.apache.syncope.client.console.annotations.BinaryPreview;
 import org.apache.syncope.client.console.annotations.ExtPage;
@@ -51,6 +53,8 @@ public class ClassPathScanImplementationLookup {
     private List<Class<? extends BaseExtPage>> extPages;
 
     private List<Class<? extends BaseExtWidget>> extWidgets;
+    
+    private List<String> basePackages = Arrays.asList("org.apache.syncope.client.console");
 
     @SuppressWarnings("unchecked")
     public void load() {
@@ -65,7 +69,11 @@ public class ClassPathScanImplementationLookup {
         scanner.addIncludeFilter(new AssignableTypeFilter(BaseExtPage.class));
         scanner.addIncludeFilter(new AssignableTypeFilter(BaseExtWidget.class));
 
-        for (BeanDefinition bd : scanner.findCandidateComponents(StringUtils.EMPTY)) {
+        Set<BeanDefinition> beanDefinitions = new HashSet<>();
+        for (String basePackage: getBasePackages()) {
+            beanDefinitions.addAll(scanner.findCandidateComponents(basePackage));
+        }
+        for (BeanDefinition bd : beanDefinitions) {
             try {
                 Class<?> clazz = ClassUtils.resolveClassName(
                         bd.getBeanClassName(), ClassUtils.getDefaultClassLoader());
@@ -161,5 +169,15 @@ public class ClassPathScanImplementationLookup {
     public List<Class<? extends BaseExtWidget>> getExtWidgetClasses() {
         return extWidgets;
     }
+    
+    /**
+     * @return basePackages for syncope class classpath scanning
+     */
+    protected List<String> getBasePackages() {
+        return basePackages;
+    }
 
+    public void setBasePackages(final List<String> basePackages) {
+        this.basePackages = basePackages;
+    }
 }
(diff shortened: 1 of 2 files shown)


ilgrosso:
Linking to SYNCOPE-926

client/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
```diff
@@ -161,5 +169,15 @@ public int compare(
...
     }
+    
+    /**
+     * @return basePackages for syncope class classpath scanning
+     */
+    protected List<String> getBasePackages() {
+        return basePackages;
+    }
 
+    public void setBasePackages(final List<String> basePackages) {
```
ilgrosso: Who is supposed to invoke this method? I would remove it.

client/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
```diff
@@ -161,5 +169,15 @@ public int compare(
     public List<Class<? extends BaseExtWidget>> getExtWidgetClasses() {
         return extWidgets;
     }
+    
+    /**
+     * @return basePackages for syncope class classpath scanning
+     */
+    protected List<String> getBasePackages() {
```
ilgrosso: Did not get what this method was for, at a first glance, then I've realized it might be used for subclasses: please mention this in the javadoc.

client/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
```diff
@@ -51,6 +53,8 @@
     private List<Class<? extends BaseExtPage>> extPages;
 
     private List<Class<? extends BaseExtWidget>> extWidgets;
+    
+    private List<String> basePackages = Arrays.asList("org.apache.syncope.client.console");
```
ilgrosso: Prior to this change, one could create classes, in its own Maven-based Syncope project, under any package, now it is forced to do it under `org.apache.syncope.client.console` but we got more startup speed.
I guess it is an acceptable trade-off.

LGTM.

core/logic/src/main/java/org/apache/syncope/core/logic/init/ClassPathScanImplementationLookup.java
```diff
@@ -74,6 +75,8 @@
     private Map<Class<? extends AccountRuleConf>, Class<? extends AccountRule>> accountRuleClasses;
 
     private Map<Class<? extends PasswordRuleConf>, Class<? extends PasswordRule>> passwordRuleClasses;
+    
+    private List<String> basePackages = Arrays.asList("org.apache.syncope.fit.core", "org.apache.syncope.core");
```
ilgrosso: Prior to this change, one could create classes, in its own Maven-based Syncope project, under any package, now it is forced to do it under `org.apache.syncope.client.console` but we got more startup speed.
I guess it is an acceptable trade-off.

LGTM.

core/logic/src/main/java/org/apache/syncope/core/logic/init/ClassPathScanImplementationLookup.java
```diff
@@ -226,5 +233,15 @@ public void load() {
 
         return passwordRuleClasses.get(passwordRuleConfClass);
     }
+    
+    /**
+     * @return basePackages for syncope class classpath scanning
+     */
+    protected List<String> getBasePackages() {
```
ilgrosso: Did not get what this method was for, at a first glance, then I've realized it might be used for subclasses: please mention this in the javadoc.

core/logic/src/main/java/org/apache/syncope/core/logic/init/ClassPathScanImplementationLookup.java
```diff
@@ -226,5 +233,15 @@ public void load() {
...
+        return basePackages;
+    }
 
+    public void setBasePackages(final List<String> basePackages) {
```
ilgrosso: Who is supposed to invoke this method? I would remove it.

ilgrosso:
The PR looks good - please apply the suggested changes and I'll merge it.

Pleae also consider that from 2.0.0.M5 we have added [ITImplementationLookup](https://github.com/apache/syncope/blob/master/fit/core-reference/src/main/java/org/apache/syncope/fit/core/reference/ITImplementationLookup.java) for integration tests and standalone, which skips classpath lookup completely.

client/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
ilgrosso: A simple `String` field should do the job.

core/logic/src/main/java/org/apache/syncope/core/logic/init/ClassPathScanImplementationLookup.java
ilgrosso: A simple `String` field should do the job.

client/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
gonzalad: I've implemented with a List<String> in order to have a similar implementation as in core.logic.init.ClassPathScanImplementationLookup.

Just tell me if you really want it to be a simple String, I'll change it.

gonzalad:
Hi Francesco, 
I've just incorporated your remarks in a second commit.
Couldn't test it with my sample apps though (I'm not at work now).
Let me know if something needs to be changed.

Thanks, 
Adrian

ilgrosso:
Thanks @gonzalad !
I have changed some small details and squashed into a single commit.

PR #37, opened 2016-09-01
Syncope 853

giacomolm:
This pull request integrates enduser test execution in maven build lifecycle. Tests are executed using  selenium, but this time an headless browser (phantomjs) is used. Since enduser is basically an angularjs application, protractor has been chosen as e2e testing framework.


ilgrosso:
Linking to SYNCOPE-853

ilgrosso:
Merged with a99618d944d27e22b6ee57fd41807613aee0563e

PR #39, opened 2016-10-31
[SYNCOPE-808] Added Netbeans Plugin to SYNCOPE 2_0_X

tmess567:
Please have a look

PR #40, opened 2016-11-21
SYNCOPE-971: Case insensitive search

gonzalad:
Global setting for activating global insensitive
search.

Insensitive search is activated when search.ignore.case
property is configured to true (default value: false).

### DIFF
--- a/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAnySearchDAO.java
+++ b/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAnySearchDAO.java
@@ -40,6 +40,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.syncope.common.lib.SyncopeConstants;
 import org.apache.syncope.common.lib.types.AnyTypeKind;
 import org.apache.syncope.common.lib.types.AttrSchemaType;
+import org.apache.syncope.core.persistence.api.dao.ConfDAO;
 import org.apache.syncope.core.provisioning.api.utils.RealmUtils;
 import org.apache.syncope.core.provisioning.api.utils.EntityUtils;
 import org.apache.syncope.core.persistence.api.dao.AnyObjectDAO;
@@ -97,6 +98,9 @@ public class JPAAnySearchDAO extends AbstractDAO<Any<?>> implements AnySearchDAO
 
     @Autowired
     private AnyUtilsFactory anyUtilsFactory;
+
+    @Autowired
+    private ConfDAO confDAO;
 
     private String getAdminRealmsFilter(
             final Set<String> adminRealms,
@@ -745,9 +749,16 @@ public class JPAAnySearchDAO extends AbstractDAO<Any<?>> implements AnySearchDAO
             final PlainSchema schema, final AttributeCond cond, final boolean not,
             final List<Object> parameters, final SearchSupport svs) {
 
-        String column = (cond instanceof AnyCond)
-                ? cond.getSchema()
-                : "' AND " + svs.fieldName(schema.getType());
+        // we activate ignoreCase only for EQ and LIKE operators
+        boolean ignoreCase = AttributeCond.Type.ILIKE == cond.getType() || AttributeCond.Type.IEQ == cond.getType();
+        boolean anyCond = cond instanceof AnyCond;
+        String column = (cond instanceof AnyCond)  ? cond.getSchema() :  svs.fieldName(schema.getType());
+        if (ignoreCase) {
+            column = "LOWER (" + column + ")";
+        }
+        if (!anyCond) {
+            column = "' AND " + column;
+        }
 
         switch (cond.getType()) {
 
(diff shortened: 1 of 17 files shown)


ilgrosso:
Hi @gonzalad thanks for your contribution!

I have noticed that you have defined an additional configuration parameter which globally states if `EQ` and `LIKE` are to be considered as case insensitive or not.

I would rather prefer introducing new operators for this purpose (`IEQ` and `ILIKE` ?).
Naturally, query builders will also need to be adjusted accordingly.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAnySearchDAO.java
```diff
@@ -769,7 +783,12 @@ private void fillAttributeQuery(final StringBuilder query, final PlainAttrValue
                     if (not) {
                         query.append(" NOT ");
                     }
-                    query.append(" LIKE ?").append(setParameter(parameters, cond.getExpression()));
+                    query.append(" LIKE ");
```
ilgrosso: This works for FIQL queries, but the search queries generated here are native.

gonzalad:
Hi @ilgrosso, the problem is I don't know how to create custom FIQL operators.

From what I've seen, we can't.
FiqlParser.operatorsMap is private & non-accessible :
https://github.com/gonzalad/cxf/blob/315e357e850407e9f3233259bab33ea765633271/rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/fiql/FiqlParser.java#L100

If you know how to implement custom fiql operators just let me know (from what I've seen in Syncope you use only native fiql operators)

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAnySearchDAO.java
gonzalad: LOWER function is defined in SQL 2003.
http://users.atw.hu/sqlnut/sqlnut2-chp-4-sect-4.html#sqlnut2-CHP-4-TABLE-9

(Sorry I didn't found a better reference)

core/persistence-jpa/src/test/java/org/apache/syncope/core/persistence/jpa/inner/AnySearchTest.java
```diff
@@ -175,6 +210,28 @@ public void searchWithNotCondition() {
     }
 
     @Test
+    public void searchCaseInsensitiveWithNotCondition() {
```
ilgrosso: This test should be replicated in [SearchITCase](https://github.com/apache/syncope/blob/2_0_X/fit/core-reference/src/test/java/org/apache/syncope/fit/core/SearchITCase.java) (like other test cases in this class)

ilgrosso:
@gonzalad let me play a bit with CXF's `FiqlParser`...

ilgrosso:
@gonzalad With https://github.com/apache/cxf/commit/5156dd256f8f59d22ea2d1ca50197f1b5b3f1788 (for CXF's `3.1.x-fixes`, which is currently used by Syncope's `2_0_X`, but I have also committed to CXF's `3.0.x-fixes` and `master`) I have pushed some changes to allow easier subclassing of `FiqlParser`.

I have also set up a [simple project](https://github.com/ilgrosso/FIQLCustomOperator) to show how to define two custom operators `=~` (case-insensitive equals) and `!~` (case-insensitive not equals).
If you would like to try it yourself, you will need to build CXF 3.1.9-SNAPSHOT from sources.

@sberyozkin WDYT?

sberyozkin:
Hi Francesco, this looks fine. 
In general, as far as the use of the CXF Search extension is concerned, it might make it tricky to reuse a CXF ODataParser (delegates to Olingo) which offers a limited OData filter support given that OData offers no such operators - however perhaps Odata functions can be supported later on if ODataParser will be used at some point of time. 
Thanks

gonzalad:
@ilgrosso : changes made, I've introduced the 2 new operators =~ and !~.
Thanks for the change in CXF fiqlParser !

2 problems in Travis CI : 
 * the build doesn't pass (cxf changes were not published I think)
 * MigrationITCase.migrateFromSyncope12 fails (I don't think this failure is caused by this changed since this test also fails in  Pull Request #39)

ilgrosso:
@gonzalad thanks for your work, the changes now look very good!

I have some troubles when merging your PR directly (as you've forked from the master branch, I think) + there are some minor issues to clean up (unused `ConfDAO`, for example).

Moreover, the CXF dependency must be upgraded to 3.1.9-SNAPSHOT in oder to get the changes pushed yesterday.
The error from `MigrationITCase` is randomly occurring, and it is probably related to some timing issue, so do not care about it.

Finally, once merged this PR, I will apply it to the `master` branch too (with no modifications) + I will port it to the `1_2_X` branch (which will instead require some intervention.

Thanks again!

gonzalad:
@ilgrosso : yes I confirm, I forked from master (I didn't know I needed to use another branch sorry !)

Are you taking care of the rest or should I do a little cleanup ? 
If I need to do some cleanup, you'll need to tell me which branch should I use for the fork (btw, master branch depends on CXF 3.2.0-SNAPSHOT)

Thanks !

ilgrosso:
>  Are you taking care of the rest or should I do a little cleanup ?

I am already at work, expect something soon :-)

ilgrosso:
I have just committed to `2_0_X` with 0e214e0046e3e9940cbe57dacf3f8bf56c1b5f9c

ilgrosso:
Before closing this PR I'll cherry-pick and adapt to `master` + rework for `1_2_X`

ilgrosso:
I have just committed to `1_2_X` with 4adf879d3ce446e1054ddc03a929e31ae25bd248

PR #44, opened 2017-01-14
Missing line-continuation symbol

mans0954:
The JAVA_OPTS in the Debian instructions appears to be missing a line-continuation symbol ('\\') at the end of the second line.


ilgrosso:
Thanks for your patch @mans0954!
Please consider [sending an ICLA](http://syncope.apache.org/contributing.html#How_do_I_become_a_contributor_or_a_committer) if you plan to further contribute to Apache Syncope.

PR #47, opened 2017-06-27
Fixes [SYNCOPE-1110], [SYNCOPE-1123]

mat-ale:
Hi,

this fixes [[SYNCOPE-1110]](https://issues.apache.org/jira/browse/SYNCOPE-1110) and [[SYNCOPE-1123]](https://issues.apache.org/jira/browse/SYNCOPE-1123).

Regards,
Matteo

### DIFF
--- a/client/enduser/src/main/java/org/apache/syncope/client/enduser/model/CustomAttribute.java
+++ b/client/enduser/src/main/java/org/apache/syncope/client/enduser/model/CustomAttribute.java
@@ -26,18 +26,18 @@ public class CustomAttribute implements Serializable {
 
     private static final long serialVersionUID = 4910266842123376686L;
 
-    private Boolean readonly;
+    private boolean readonly;
 
     private List<String> defaultValues = new ArrayList<>();
 
     public CustomAttribute() {
     }
 
-    public Boolean getReadonly() {
+    public boolean isReadonly() {
         return readonly;
     }
 
-    public void setReadonly(final Boolean readonly) {
+    public void setReadonly(final boolean readonly) {
         this.readonly = readonly;
     }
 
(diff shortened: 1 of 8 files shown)


client/enduser/src/main/java/org/apache/syncope/client/enduser/model/CustomAttribute.java
```diff
@@ -26,18 +26,18 @@
...
-    private Boolean readonly;
+    private boolean readonly;
 
     private List<String> defaultValues = new ArrayList<>();
 
     public CustomAttribute() {
     }
 
-    public Boolean getReadonly() {
+    public boolean getReadonly() {
```
ilgrosso: This should be `isReadonly()`

client/enduser/src/main/java/org/apache/syncope/client/enduser/resources/UserSelfReadResource.java
```diff
@@ -137,12 +137,13 @@ public void writeData(final Attributes attributes) throws IOException {
     private void customizeAttrs(final Set<AttrTO> attrs,
             final Map<String, CustomAttribute> customForm) {
 
-        CollectionUtils.filter(attrs, new Predicate<AttrTO>() {
-
-            @Override
-            public boolean evaluate(final AttrTO attr) {
-                return customForm.containsKey(attr.getSchema());
+        Set<AttrTO> attrsToAdd = new HashSet<>();
```
ilgrosso: You could use `CollectionUtils.collect()` here.

fit/enduser-reference/src/test/resources/protractor-conf.js
```diff
@@ -19,17 +19,17 @@
 exports.config = {
   directConnect: false,
   capabilities: {
-    'browserName': 'phantomjs',
+    'browserName': 'chrome',
```
ilgrosso: Bad: restore `phantomjs` here!

andrea-patricelli:
@mat-ale we have almost finished. Please put a check to call  completeUserObject only if cuastomForm is not null and not empty and add user virtual attributes management in  completeUserObject.

PR #59, opened 2017-08-18
SYNCOPE-1197 SAML label inserted and combo moved below login button

loredicola:
spostato menù saml e aggiunta stringa informativa


ilgrosso:
Thanks @loredicola for this PR

PR #62, opened 2017-08-25
SYNCOPE-1196 add a specific form for binary and encrypted paramiters …

loredicola:
…type and set a custom label in paramiters list


ilgrosso:
Thanks @loredicola!

PR #67, opened 2018-02-20
Modifiying documentation for netbeans IDE plugin

rohanjulka19:
When I downloaded the Netbeans plugin it was in zip format so I was not aware of what to do(silly me) . So I added a note which may help some users.

### DIFF
--- a/src/main/asciidoc/getting-started/obtain.adoc
+++ b/src/main/asciidoc/getting-started/obtain.adoc
@@ -727,6 +727,10 @@ The Netbeans IDE plugin allows remote management of notification e-mail and repo
 After http://syncope.apache.org/downloads.html[download^], start the most recent Netbeans IDE then go to 
 `Tools > Plugins`:
 
+****
+Note: :: If the downloaded plugin is in zip format then change its extension to .nbm otherwise you won't be able to use it with Netbeans
+****
+
 image::netbeans01.png[netbeans01]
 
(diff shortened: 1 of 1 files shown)


ilgrosso:
There is no "If the downloaded plugin is in zip format", it *is* in zip format, please adjust.

Also, wrap the note into an asciidoctor's `[TIP]` block,

rohanjulka19:
Yes sir

rohanjulka19:
Is this ok 
"The downloaded plugin is in zip format , change its extension to .nbm otherwise you won't be able to use it with NetBeans"

ilgrosso:
You don't need to "change its extension", you need to unzip it.

rohanjulka19:
I unzipped the file  but was not able to add it as a plugin .

ilgrosso:
Ah, correct it's .nbm indeed: why on earth you have downloaded as ZIP?

ilgrosso:
Please close this PR, I don't see reasons to merge it.

rohanjulka19:
I went to the link specified in the documentation and it downloads a zip file http://www.apache.org/dyn/closer.lua/syncope/2.0.7/syncope-ide-netbeans-2.0.7.nbm

rohanjulka19:
Sorry for wasting your time

ilgrosso:
np

PR #70, opened 2018-04-12
[SYNCOPE-1301] fixed

IsurangaPerera:


core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAccessTokenDAO.java
```diff
@@ -115,6 +115,16 @@ public AccessToken save(final AccessToken accessToken) {
         return entityManager().merge(accessToken);
     }
 
+    @Override
+    @Transactional(rollbackFor = Throwable.class)
```
ilgrosso: Please remove this annotation, it is redundant..

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAccessTokenDAO.java
```diff
@@ -115,6 +115,16 @@ public AccessToken save(final AccessToken accessToken) {
...
+    public void merge(final AccessToken accessToken) {
+        AccessToken existing = findByOwner(accessToken.getOwner());
+        if (existing != null) {
+            entityManager().remove(existing);
+        }
+        entityManager().persist(accessToken);
```
ilgrosso: replace with `save(accessToken)` and make this method return `AccessToken` as `save()` does.

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/AccessTokenDAO.java
```diff
@@ -34,6 +34,8 @@
 
     AccessToken save(AccessToken accessToken);
 
+    void merge(AccessToken accessToken);
```
ilgrosso: Change it into
```
AccessToken merge(AccessToken accessToken);
```

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/AccessTokenDataBinderImpl.java
```diff
@@ -135,11 +135,7 @@
                 accessToken.setAuthorities(authorities);
             }
 
-            accessTokenDAO.save(accessToken);
-        }
-
-        if (replaceExisting && existing != null) {
-            accessTokenDAO.delete(existing);
+            accessTokenDAO.merge(accessToken);
```
ilgrosso: what about the `replaceExistingFlag`? You just removed it!

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/JPAAccessToken.java
```diff
@@ -44,7 +44,7 @@
     @Temporal(TemporalType.TIMESTAMP)
     private Date expiryTime;
 
-    @Column(nullable = true)
+    @Column(unique = true)
```
ilgrosso: As said during the initial mail exchange, I am not sure of the reason why owner is currently set as nullable, and I don't see enough reasons to change it, please revert.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/AccessTokenDataBinderImpl.java
IsurangaPerera: yes, merge function simply ensure the uniqueness constraint by checking if the token exists and replace it if so. otherwise, it will create a new register. so no need to perform this check here

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/JPAAccessToken.java
IsurangaPerera: I think in JPA it is redundant to use nullable = true this is the default scenario or all columns in JPA. Should I still change it?

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/AccessTokenDataBinderImpl.java
ilgrosso: `replaceExisting` is an argument of `AccesstTokenDataBinderImpl#create` that your change would simply ignore.
Not good: it must be checked.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/JPAAccessToken.java
ilgrosso: Just revert, I know `nullable = true` is the default, but I don't see reasons to change.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/AccessTokenDataBinderImpl.java
IsurangaPerera: It is not completely ignored. it is checked by the second if condition as before

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/AccessTokenDAO.java
IsurangaPerera: Shall I keep the method name same(merge) or change it?

ilgrosso: maybe `replace()` makes more sense

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAccessTokenDAO.java
ilgrosso: Also, replace `entityManager().remove(existing)` with `delete(existing)`, thanks.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/AccessTokenDataBinderImpl.java
ilgrosso: So, the current implementation will replace the token only if `replaceExisting` is true; with your change, the token will be replaced anyway, because `JPAAccessTokenDAO#merge` will do

```
if (existing != null) {
            entityManager().remove(existing);
        }
```
while the current code does
```
        if (replaceExisting && existing != null) {
            accessTokenDAO.delete(existing);
        }
```

IsurangaPerera: exactly. this is to assure that there can be only one access token associated with a particular user at an any given time and also a workaround to overcome thread not safe problem without affecting the performance(without using serialized isolation)

ilgrosso: ...but there might be cases where such constraint should not apply (see the `@Column` annotation we've been discussing).

ilgrosso: To me, it should be enough to pass the `replaceExisting` flag to the new `AccessTokenDAO#replace` method, and behave accordingly.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAccessTokenDAO.java
```diff
@@ -115,6 +115,16 @@ public AccessToken save(final AccessToken accessToken) {
...
     }
 
+    @Override
+    public AccessToken replace(final AccessToken accessToken) {
+        AccessToken existing = findByOwner(accessToken.getOwner());
+        if (existing != null) {
+            delete(existing.getKey());
+        }
+
+        return entityManager().merge(accessToken);
```
ilgrosso: replace `entityManager().merge(accessToken)` with `save(accesToken)`

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/AccessTokenDAO.java
```diff
@@ -34,6 +34,8 @@
 
     AccessToken save(AccessToken accessToken);
 
+    AccessToken replace(AccessToken accessToken);
```
ilgrosso: This should be 

```
AccessToken replace(boolean replaceExisting, AccessToken accessToken);
```

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAccessTokenDAO.java
```diff
@@ -115,6 +115,16 @@ public AccessToken save(final AccessToken accessToken) {
...
     }
 
+    @Override
+    public AccessToken replace(final AccessToken accessToken) {
+        AccessToken existing = findByOwner(accessToken.getOwner());
+        if (existing != null) {
+            delete(existing.getKey());
+        }
+
+        return save(accessToken);
```
ilgrosso: `return replaceExisting ? save(accessToken) : existing;`

IsurangaPerera:
@ilgrosso without enforcing the UNIQUE constraint on the owner as we discussed in the mail I can't see a valid reason for merge these changes

ilgrosso:
@IsurangaPerera the point is that the existence of `replaceFlag` (I did not realize that at first) is what prevents to enforce the UNIQUE constraint.

IsurangaPerera:
@ilgrosso  Can I know the business rule associated with the access token. Can one user has more than one access token in an any given scenario?

ilgrosso:
@IsurangaPerera I don't remember the details, but what I can see from the source three is that `AccessTokenDataBinderImpl#create` is invoked in two places:

1. https://github.com/apache/syncope/blob/master/core/logic/src/main/java/org/apache/syncope/core/logic/AccessTokenLogic.java#L84 (plain login)
1. https://github.com/apache/syncope/blob/master/ext/saml2sp/logic/src/main/java/org/apache/syncope/core/logic/SAML2SPLogic.java#L549 (SAML 2.0 login)

The former provides `replaceExisting` as `false`, the latter as `true`.

From the code in `AccessTokenDataBinderImpl#create` I can see that:

* for plain login, JWT is generated only at first invocation
* for SAML 2.0 login, JWT is generated at every invocation, and existing JWT is replaced if existing

Moreover, I cannot recall exactly why the UNIQUE constraint is not imposed to AccessToken's `owner`.

ilgrosso:
Ok, let's try this way: please re-introduce the UNIQUE constraint but rework the whole `AccessTokenDataBinderImpl#create` to comply with the two invocation scenarios as reported above.

Do you think it could be feasible?

IsurangaPerera:
@ilgrosso As I understand in SAML SP logic always replaced. So even when we logged, as usual, the access token may changed by SAML SP. So I can understand the importance of what replaceExisitng flag does. After imposing the UNIQUE constraint as in my implementation replaceExisting == true works as expected(always).But sometimes even if the flag is false the token may be replaced (scenario discussed in mail thread). But this is only when the same user tries to log in at the same time & thread not safe problem aise. Anyway this approach is far better than using locks which causes performance drop and this is a rare case as well. What do you think?

ilgrosso:
@IsurangaPerera the only effect of UNIQUE constraint will be that a validation error is raised whenever a second AccessToken is created with same owner of an existing one.
Are you simply proposing to add such constraint and leave the rest of the code as is?

IsurangaPerera:
A validation error will be raised in the worst case(As I see the thread safe problem won't just affect the login as we discussed Saml as well. So Unique constraint will only prevent those scenarios. the advantage is the validation error is the worst case. It won't negatively affect the flow of the system like in thrad safe problem(Users being blocked from login)

ilgrosso:
Ok so...
> Are you simply proposing to add such constraint and leave the rest of the code as is?

e.g. to change again this PR by reverting all changes and adding the UNIQUE constraint?

IsurangaPerera:
Yes, What if we revert it to the initial state?

ilgrosso:
I would stick with minimal changes, so this PR should only change `JPAAccessToken`.

ilgrosso:
> That way replace won't work since it saves first (2 tokens exist.violate UNIQUE constraint) and deletes next. That way it will only delete the existing one which results in no token at all.

You are right, but as you can see, in the [current implementation](https://github.com/apache/syncope/blob/master/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/AccessTokenDataBinderImpl.java#L138-L142) save and delete do not happen under the same conditions.

IsurangaPerera:
Yes, that's why we have to see it from a different perspective.
The initial state doesn't change the logic of replaceExisting == true
But it partially affects replaceExisting == False (This affects only if the thread safe problems occur). This means when the issue syncope-1301 occurs instead of creating a new token for the same user this will replace the existing one (rare situation)

PR #72, opened 2018-04-26
Syncope 1306

DimaAy:
This pull request is to fix the issue [SYNCOPE-1306] that is about adding a date value without a conversion pattern


DimaAy:
sorry

PR #74, opened 2018-05-02
[SYNCOPE-1270] implementation for OpenID Connect for Admin Console and Enduser

DimaAy:


ilgrosso:
Thanks @DimaAy the code looks almost complete, and it's also working.

I would just make some enhancements to the Admin Console management:

1. When adding / editing an OIDC provider, validate all values that are expected to be URIs *before* saving
1. Move the `Has Discovery Document` flag after `Issuer`
1. When editing an existing ODIC provider, make all the second step's fields as read-only, and show them all despite the Discovery Document was used during create

Side note: I see there is no yet support for [Logout](https://medium.com/@robert.broeckelmann/openid-connect-logout-eccc73df758f), which is provided at least by [Azure](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/development/ad-fs-logout-openid-connect), maybe you can submit a new PR for that after the current ones gets merged.

ilgrosso:
Thanks @DimaAy for your changes, they look fine!
Just squash your 3 commits into 1 and I'll gladly merge this PR, thanks.

PR #84, opened 2018-10-23
SYNCOPE-1387 fix class cast exception, remove redundant actions call

AlexTsvetkov:
https://issues.apache.org/jira/browse/SYNCOPE-1387


ilgrosso:
That call is *not* redundant, it is invoking the `after()` method with any defined push actions.
Please close this PR.

PR #85, opened 2018-10-23
Revert: SYNCOPE-1387 fix class cast exception

AlexTsvetkov:
https://issues.apache.org/jira/projects/SYNCOPE/issues/SYNCOPE-1387


ilgrosso:
@AlexTsvetkov I have reviewed the code in `DefaultRealmPullResultHandler` and found that you were right about removing the additional `after()` calls - as proposed in #84 .

This because the `link()` method eventually calls `update()` - thus I have tried to clean up a bit the call hierarchy in 5e854e117594da3a1ba78c0af8d53d89a66a9c5a

Please close this PR as well, sorry for the noise; at least, now the issue should be fixed.

For future contributions, please submit an [ICLA](http://syncope.apache.org/contributing#How_do_I_become_a_contributor_or_a_committer.3F), thanks.

PR #86, opened 2018-11-01
SYNCOPE-1390: Pull Realms: pull task with Unmatching Rules: PROVISION…

AlexTsvetkov:
… shouldn't create propagation task


ilgrosso:
Thanks for your contribution @AlexTsvetkov - please do not forget to submit your ICLA.

PR #88, opened 2018-11-05
SYNCOPE-1393 jexl function fullPath2Dn(final String fullPath, final S…

AlexTsvetkov:
…tring attr, final String prefix) should return empty string for ROOT realm


ilgrosso:
Thanks @AlexTsvetkov

PR #96, opened 2019-02-14
Enable security-related HTTP headers in the console

coheigea:
It's good security practice to enable the HTTP headers included in the patch. See here:

https://ci.apache.org/projects/wicket/guide/8.x/single.html#_external_security_checks

I created the PR against 2.1.x, as I'm not sure what state master is currently in.


ilgrosso:
LGTM

I would have actually preferred if there were some constants around for `X-Frame-Options`, `X-XSS-Protection` etc. (from dependencies I mean, not in Syncope) but it seems that only Spring Security has something (as [1](https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/server/header/XFrameOptionsServerHttpHeadersWriter.java#L28) or [2](https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/server/header/XXssProtectionServerHttpHeadersWriter.java#L29)).

Please go ahead and merge.

ilgrosso:
> I created the PR against 2.1.x, as I'm not sure what state master is currently in.

Don't worry: once merged, I'll take care of cherry-picking against `2_0_X` and `master` branches.

PR #98, opened 2019-02-28
[SYNCOPE-1439] Fixed user membership attributes not updating from Admin Console

mat-ale:
Fix for issue [SYNCOPE-1439](https://issues.apache.org/jira/browse/SYNCOPE-1439) on version `2.0.x`.

### DIFF
--- a/client/console/src/main/java/org/apache/syncope/client/console/wizards/any/PlainAttrs.java
+++ b/client/console/src/main/java/org/apache/syncope/client/console/wizards/any/PlainAttrs.java
@@ -18,6 +18,7 @@
  */
 package org.apache.syncope.client.console.wizards.any;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


client/console/src/main/java/org/apache/syncope/client/console/wizards/any/PlainAttrs.java
```diff
@@ -112,7 +115,7 @@
 
             @Override
             public WebMarkupContainer getPanel(final String panelId) {
-                return new PlainSchemas(panelId, schemas, attrTOs);
+                return new PlainSchemasSimple(panelId, schemas, attrTOs);
```
ilgrosso: Why not name `PlainSchemasSimple` as `PlainSchemasOwn` (considering that the other one is named `PlainSchemasMemberships`)?

client/console/src/main/java/org/apache/syncope/client/console/wizards/any/PlainAttrs.java
```diff
@@ -363,15 +375,91 @@ protected PageReference getPageReference() {
         return panel;
     }
 
-    public class PlainSchemas extends Schemas {
+    public class PlainSchemasMemberships extends PlainSchemas<AttributableTO> {
```
ilgrosso: Any point to have this class and its sister class as public? Why not `protected static`, or at least `protected` if needed?

client/console/src/main/java/org/apache/syncope/client/console/wizards/any/PlainAttrs.java
mat-ale: Good point, I'll do that thanks!

client/console/src/main/java/org/apache/syncope/client/console/wizards/any/PlainAttrs.java
mat-ale: Yes `protected` seems better!

PR #99, opened 2019-03-11
SYNCOPE-1448 add synchronized block for bean loading/register section

longstone:
Issue [SYNCOPE-1448](https://issues.apache.org/jira/projects/SYNCOPE/issues/SYNCOPE-1448)

When two threads (T1,T2) arrive the section `if (ApplicationContextProvider.getBeanFactory().containsSingleton(ruleClass.getName())) {`at the same time they will both assume that there is no bean, which is right so far.

T1 register the bean as it should here `ApplicationContextProvider.getBeanFactory().registerSingleton(ruleClass.getName(), rule);`. As soon as T2does the same, an exception is thrown. 

Solution:

When T1 and T2 found that a given bean is not there, they'll hit a synchronized section with double checking. This will prevent the duplicated creation and registration of a bean.

### DIFF
--- a/core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
+++ b/core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
@@ -104,10 +104,19 @@ public final class ImplementationManager {
                         rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
                                 getSingleton(ruleClass.getName());
                     } else {
-                        rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
-                                createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                        ApplicationContextProvider.getBeanFactory().
-                                registerSingleton(ruleClass.getName(), rule);
+                        // to avoid concurrency issues
+                        synchronized (ImplementationManager.class) {
+                            // double check for performance reasons
+                            if (ApplicationContextProvider.getBeanFactory().containsSingleton(ruleClass.getName())) {
+                                rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
+                                        getSingleton(ruleClass.getName());
+                            } else {
+                                rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
+                                        createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
+                                ApplicationContextProvider.getBeanFactory().
+                                        registerSingleton(ruleClass.getName(), rule);
+                            }
+                        }
                     }
                     rule.setConf(ruleConf);
                 }
@@ -138,10 +147,19 @@ public final class ImplementationManager {
                         rule = (PasswordRule) ApplicationContextProvider.getBeanFactory().
                                 getSingleton(ruleClass.getName());
                     } else {
-                        rule = (PasswordRule) ApplicationContextProvider.getBeanFactory().
-                                createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                        ApplicationContextProvider.getBeanFactory().
-                                registerSingleton(ruleClass.getName(), rule);
+                        // to avoid concurrency issues
+                        synchronized (ImplementationManager.class) {
+                            // double check for performance reasons
+                            if (ApplicationContextProvider.getBeanFactory().containsSingleton(ruleClass.getName())) {
+                                rule = (PasswordRule) ApplicationContextProvider.getBeanFactory().
+                                        getSingleton(ruleClass.getName());
+                            } else {
+                                rule = (PasswordRule) ApplicationContextProvider.getBeanFactory().
+                                        createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
+                                ApplicationContextProvider.getBeanFactory().
+                                        registerSingleton(ruleClass.getName(), rule);
+                            }
+                        }
                     }
                     rule.setConf(ruleConf);
                 }
@@ -173,10 +191,19 @@ public final class ImplementationManager {
                         rule = (PullCorrelationRule) ApplicationContextProvider.getBeanFactory().
                                 getSingleton(ruleClass.getName());
                     } else {
-                        rule = (PullCorrelationRule) ApplicationContextProvider.getBeanFactory().
-                                createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                        ApplicationContextProvider.getBeanFactory().
-                                registerSingleton(ruleClass.getName(), rule);
+                        // to avoid concurrency issues
+                        synchronized (ImplementationManager.class) {
+                            // double check for performance reasons
+                            if (ApplicationContextProvider.getBeanFactory().containsSingleton(ruleClass.getName())) {
+                                rule = (PullCorrelationRule) ApplicationContextProvider.getBeanFactory().
+                                        getSingleton(ruleClass.getName());
+                            } else {
+                                rule = (PullCorrelationRule) ApplicationContextProvider.getBeanFactory().
+                                        createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
+                                ApplicationContextProvider.getBeanFactory().
+                                        registerSingleton(ruleClass.getName(), rule);
+                            }
+                        }
                     }
                     rule.setConf(ruleConf);
                 }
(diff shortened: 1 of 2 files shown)


ilgrosso:
Hi @longstone , thanks for this PR.
The change you are proposing looks generally fine to me - I would like to double-check before merging, anyway.

Meanwhile, please:

1. open an issue on [JIRA](https://issues.apache.org/jira/projects/SYNCOPE/)
1. change the commit log message to include the issue number (something like as `[SYNCOPE-XXXX] <the current log message>`
1. submit your ICLA, as explained [here](http://syncope.apache.org/contributing#How_do_I_become_a_contributor_or_a_committer.3F)

Thanks!

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -104,10 +104,19 @@
                         rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
                                 getSingleton(ruleClass.getName());
                     } else {
-                        rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
-                                createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                        ApplicationContextProvider.getBeanFactory().
-                                registerSingleton(ruleClass.getName(), rule);
+                        // to avoid concurrency issues
+                        synchronized (ImplementationManager.class) {
```
ilgrosso: Why to synchronize on a class object? You could synchronize on `LOG`, for instance.

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -104,10 +104,19 @@
...
+                            // double check for performance reasons
+                            if (ApplicationContextProvider.getBeanFactory().containsSingleton(ruleClass.getName())) {
```
ilgrosso: Why to repeat this check, which was already performed just a few lines above?
You can refactor the whole block as follows, simplify the logic and avoid duplicates:

```java
                // fetch (or create) rule
                synchronized (LOG) {
                    if (ApplicationContextProvider.getBeanFactory().containsSingleton(ruleClass.getName())) {
                        rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
                                getSingleton(ruleClass.getName());
                    } else {
                                rule = (AccountRule) ApplicationContextProvider.getBeanFactory().
                                        createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
                                ApplicationContextProvider.getBeanFactory().
                                        registerSingleton(ruleClass.getName(), rule);
                        }
                }
```

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -138,10 +147,19 @@
                         rule = (PasswordRule) ApplicationContextProvider.getBeanFactory().
                                 getSingleton(ruleClass.getName());
                     } else {
-                        rule = (PasswordRule) ApplicationContextProvider.getBeanFactory().
-                                createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                        ApplicationContextProvider.getBeanFactory().
-                                registerSingleton(ruleClass.getName(), rule);
+                        // to avoid concurrency issues
```
ilgrosso: same as above

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -173,10 +191,19 @@
                         rule = (PullCorrelationRule) ApplicationContextProvider.getBeanFactory().
                                 getSingleton(ruleClass.getName());
                     } else {
-                        rule = (PullCorrelationRule) ApplicationContextProvider.getBeanFactory().
-                                createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                        ApplicationContextProvider.getBeanFactory().
-                                registerSingleton(ruleClass.getName(), rule);
+                        // to avoid concurrency issues
```
ilgrosso: same as above

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -208,10 +235,19 @@
                         rule = (PushCorrelationRule) ApplicationContextProvider.getBeanFactory().
                                 getSingleton(ruleClass.getName());
                     } else {
-                        rule = (PushCorrelationRule) ApplicationContextProvider.getBeanFactory().
-                                createBean(ruleClass, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                        ApplicationContextProvider.getBeanFactory().
-                                registerSingleton(ruleClass.getName(), rule);
+                        // to avoid concurrency issues
```
ilgrosso: same as above

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
longstone: assumption: 
* synchronized is expensive and therefore I want to use it as less as possible
* I want to register the bean exactly one time

If you do it like this, every time a rule is loaded its synchronized. This means it's in a sequential mode. and we do it exclusively every time. This works but is also a lack of performance

If you do it with two checks,  you'll get the first check as usual. the will only be executed from the first few callers. so, only they pay the price for synchronized.

Some references (this is more for a singleton but same patterns apply) #5: https://www.geeksforgeeks.org/java-singleton-design-pattern-practices-examples/
also here https://stackoverflow.com/questions/11165852/java-singleton-and-synchronization

ilgrosso: So the logic is:

1. check if there is a singleton bean
    1. if so, take it
    1. if not, synchronize and check again
          1. if so, take it
          1. if not, create it

Also, did you take into account that [DefaultSingletonBeanRegistry#containsSingleton](https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultSingletonBeanRegistry.java#L285) is backed by a `ConcurrentHashMap`?

About the provided references, I am not convinced they apply to the case as above.

ilgrosso:
As noted above, [DefaultSingletonBeanRegistry#containsSingleton](https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultSingletonBeanRegistry.java#L285) is backed by a `ConcurrentHashMap`, the same map used by [DefaultSingletonBeanRegistry#registerSingleton](https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultSingletonBeanRegistry.java#L114).
As you can see there, the code already does synchronization on that map and raises an `IllegalStateException` in case a ben with the given name is already present.

Given the implementation in Spring, I think that SYNCOPE-1488 should be fixed on Syncope in a completely different way, e.g. simply catching and logging the `IllegalStateException`, as all the needed concurrency control is - correctly BTW - performed by Spring code.

longstone:
I faced the issue when using syncope and therefore the hashmap is not securing the registration of the beans. Maybe I messed up with version. 
However, this part is not Threadsafe and a exception is thrown when registering a bean which is already registered.
Give me some time. Maybe weekend.

longstone:
I submitted a test. Guarantees atomic operations on the HashMap. The fix I'm supposing is handling it on a different Level. 
Up to you to decide on how to go further.

ilgrosso:
Thanks, I'll take a look. Meanwhile, please add the license header to the new class, otherwise checks will fail anyway.

ilgrosso:
@longstone I attempted to run your add test against `master` branch, and it failed as expected, good.

Then I applied my suggested fix (catch and log `IllegalStateException`) and now it passes: I have committed such alternative fix into the [SYNCOPE-1448_alternative-fix](https://github.com/apache/syncope/tree/SYNCOPE-1448_alternative-fix) branch, if you want to take a look.

longstone:
@ilgrosso you can do it also like this. I have following thoughts about this solution
* you're first few threads will may not have the "same" bean, as they instantiate it for it own -> _I can't judge what possible side effects may be_
* we (and the customers) will find a WARN message in the log -> _would be great if can be changed to info or debug_?

ilgrosso:
> you're first few threads will may not have the "same" bean, as they instantiate it for it own -> I can't judge what possible side effects may be

you are right, please see the updated class in the `SYNCOPE-1448_alternative-fix` branch.

> we (and the customers) will find a WARN message in the log -> would be great if can be changed to info or debug?

To me, `warn` makes more sense, but I'll go with `debug` if you ask so.

ilgrosso:
@longstone are you ok with the alternative fix, after last changes I made yesterday? Shall I go ahead and merge that to `master` and `2_1_X`?

ilgrosso:
no feedback for 6 days, I went ahead and fixed via 50469a15b51b99e6b1e52a05626f96e0b9ce6cdd

PR #101, opened 2019-03-27
[SYNCOPE-1453] MappingItem with "mustChangePassword" field cannot be provisioned and updated during import

DmitriyBrashevets:


core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -349,6 +351,15 @@ public PropagationByResource update(final User toBeUpdated, final UserPatch user
 
         if (userPatch.getMustChangePassword() != null) {
             user.setMustChangePassword(userPatch.getMustChangePassword().getValue());
+            Set<ExternalResource> externalResources = anyUtils.getAllResources(toBeUpdated);
```
ilgrosso: The same code for adding resources to the `propByRes` instance in case such resources contain the corresponding mapping item should also performed in the `create()` method of this class - here being the `update()` method.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -668,4 +679,13 @@ public UserTO getUserTO(final String key) {
         return getUserTO(userDAO.authFind(key), true);
     }
 
+    private boolean isMustChangePwdMappingItemConfigured(Mapping mapping) {
```
ilgrosso: Use `IterableUtils#matchesAny` as in other methods of the same class - and in the whole code, at least in the `2_0_X` branch.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/utils/ConnObjectUtils.java
```diff
@@ -266,7 +266,6 @@ public RealmTO getRealmTO(final ConnectorObject obj, final PullTask task, final
                     }
 
                     updatedUser.setSecurityQuestion(updatedUser.getSecurityQuestion());
-                    updatedUser.setMustChangePassword(originalUser.isMustChangePassword());
```
ilgrosso: Why is this line removed? Are you sure this works either when `mustChangePassword` is mapped and when is not mapped?

DmitriyBrashevets: I will check this

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
DmitriyBrashevets: All resources during the create operation are being added here https://github.com/apache/syncope/blob/85aadb3cd826d34e984540990187552437aa9236/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/AbstractAnyDataBinder.java#L504. Finally the propByRes are populated from saved user entity at this place https://github.com/apache/syncope/blob/85aadb3cd826d34e984540990187552437aa9236/core/workflow-java/src/main/java/org/apache/syncope/core/workflow/java/DefaultUserWorkflowAdapter.java#L85. After this final provisioning into remote storages is happening here https://github.com/apache/syncope/blob/85aadb3cd826d34e984540990187552437aa9236/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/DefaultUserProvisioningManager.java#L92

I assume that nothing should be added for `create` operation. Am I missing something? Please, correct me if I'm wrong.

DmitriyBrashevets: Note that I'm merging into 2_0_X branch and haven't explored other branches.

ilgrosso: No, you're right: all resources are already in `propByRes` - resolving this.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -668,4 +679,13 @@ public UserTO getUserTO(final String key) {
...
+        return IterableUtils.matchesAny(mapping.getItems(), new Predicate<MappingItem>() {
+            @Override
+            public boolean evaluate(MappingItem mappingItem) {
```
ilgrosso: `final` is missing here, build will fail

DmitriyBrashevets: added it.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
ilgrosso: `final` also missing here

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/utils/ConnObjectUtils.java
DmitriyBrashevets: 1. this line always adds the value to updatedUser from originalUser. It means that `mustChangePassword` flag will never be updated during pull operation.
2. Also, it is removed to correctly calculate the UserPatch. Otherwise, `org.apache.syncope.common.lib.patch.UserPatch#mustChangePassword` is calculated incorrectly.

DmitriyBrashevets:
Rebased on 2_0_X since you've added the commit https://github.com/apache/syncope/commit/85aadb3cd826d34e984540990187552437aa9236 and I had the conflicts

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/utils/ConnObjectUtils.java
ilgrosso: I agree that the current line is not correct, but the actual fix is to just remove it, the actual fix is to avoid that line in case `mustChangePassword` is mapped.

In order to do so, I would propose the following changes:

* move `UserDataBinderImpl#isMustChangePwdMappingItemConfigured` to `MappingUtils` as follows:
```java
    public static boolean hasMustChangePassword(final Provision provision) {
        return provision != null && provision.getMapping() != null
                && IterableUtils.matchesAny(provision.getMapping().getItems(), new Predicate<MappingItem>() {

                    @Override
                    public boolean evaluate(final MappingItem mappingItem) {
                        return "mustChangePassword".equals(mappingItem.getIntAttrName());
                    }
                });
    }
```
* adjust (and simplify) the code you added in `UserDataBinderImpl#update`
* in `ConnObjectUtils`, replace
```java
updatedUser.setMustChangePassword(originalUser.isMustChangePassword());
```
with
```java
if (!MappingUtils.hasMustChangePassword(provision)) {
  updatedUser.setMustChangePassword(originalUser.isMustChangePassword());
}
```

WDYT?

DmitriyBrashevets: Yes, you're right. If mustChangePassword is not mapped and originalUser has mustChangePwd field set to true, than the value will be overwritten and set to `false`. Agree with you. Will apply your code.

DmitriyBrashevets: updated the code as you suggested

ilgrosso:
Thank you very much @DmitriyBrashevets !

Also merged to:
* `2_1_X` via ac58796c8fc10b484389b63abca1182297768224
* `master` via 37c6aacd0dbfccdea7e98c95c94aea882b563d1b

(with some minor modifications)

PR #103, opened 2019-06-10
[SYNCOPE-163] Started to implement the persistence layer for AM

mdisabatino:


mdisabatino:
This first commit aims to set the the first part of the persistence layer for the AM module. I have implemented the interfaces for authentication policies, modules and chains. I provided for each entity the possibility of adding implementations.
I created a pull request because there are some interfaces that still need to be implemented / modified (missing dependency, new module, etc etc ), so before continuing I would like to have feedback from you.

ilgrosso:
Closed in favor of the work being done in the `SYNCOPE-1631-1` branch.

PR #104, opened 2019-07-01
Added database name to JDBC URL

rshoesmi:


ilgrosso:
Hi @rshoesmi thanks for your contribution.
Can you please provide some background? Why do you think such change is needed?

rshoesmi:
Sycope would not start up after switching to the file based H2 database as described in the documentation. Adding the database name to the jdbc url fixed the problem.

ilgrosso:
Sorry, I cannot replicate the issue you describe: following the docs as they are right now worked fine for me (on Linux); the following files were created in my home directory, as expected:

* syncopedb.mv.db
* syncopedb.trace.db

With your suggested changes, the following files, instead, again as expected:

* syncopedb:syncopedb.mv.db
* syncopedb:syncopedb.trace.db

Are you on a different platform than Linux?

Anyway, can you please report the error you get when following the current docs?

rshoesmi:
This was on macOS Mojave 10.14.5.

I deleted all the database files to start fresh and it's working now. Perhaps they got corrupted somehow.

ilgrosso:
Ah ok, glad it works now for you anyway!

PR #105, opened 2019-09-03
SYNCOPE-1483: maven-enforcer-plugin: API incompatibility fails the build

mmoayyed:
See https://issues.apache.org/jira/browse/SYNCOPE-1483

Switching to 3.0.0.M2 version of plugin allows for a successful build.


ilgrosso:
LGTM thanks @mmoayyed !

PR #107, opened 2019-09-04
SYNCOPE-1332: Java 11 language features; use factory calls for immutable collections

mmoayyed:
See https://issues.apache.org/jira/browse/SYNCOPE-1332

> Java 9 came with static methods on the List, Set and Map interfaces for creating unmodifiable instances of those collections. Till now, to create immutable collection, programmers had to construct it, store it in a local variable, fill with add method and wrap into a unmodifiable collection.

There are many other references to unmodifiableList, unmodifiableSet, etc. However the conversion isn't quite straight forward to begin with, and due to API incompatibility, the inner collection must turn into an array before it be wrapped in a factory call.

Example:

```
return Collections.unmodifiableSet(values);
```

Which could turn into:

```
return Set.of(values.toArray(new Object[] {}));
```

Not sure if there is any value in covering those, so I skipped them for now.


client/idrepo/console/src/main/java/org/apache/syncope/client/console/SyncopeWebApplication.java
```diff
@@ -84,10 +84,9 @@
 
     private static final String CONSOLE_PROPERTIES = "console.properties";
 
-    public static final List<Locale> SUPPORTED_LOCALES = Collections.unmodifiableList(Arrays.asList(
-            new Locale[] {
-                Locale.ENGLISH, Locale.ITALIAN, new Locale("pt", "BR"), new Locale("ru"), Locale.JAPANESE
-            }));
+    public static final List<Locale> SUPPORTED_LOCALES = List.of(new Locale[]{
```
ilgrosso: why not simply

```List.of(Locale.ENGLISH, Locale.ITALIAN, ...);```

?

**Comment by ilgrosso** [comment]
@mmoayyed The codebase is full of `Arrays.asList()` or similar, thanks for cleaning this up!

**Comment by mmoayyed** [comment]
Sure thing. Updated the pull to catch all.

**Comment by ilgrosso** [comment]
@mmoayyed the changes included here seem to fail the build - see https://travis-ci.org/apache/syncope/builds/580651810

File: client/idm/console/src/main/java/org/apache/syncope/client/console/wizards/resources/AbstractMappingPanel.java [hunk 2]
```diff
@@ -228,7 +227,7 @@ protected void populateItem(final ListItem<ItemTO> item) {
                         "mandatoryCondition",
                         new PropertyModel<>(itemTO, "mandatoryCondition"));
                 mandatory.hideLabel();
-                mandatory.setChoices(Arrays.asList(new String[] { "true", "false" }));
+                mandatory.setChoices(List.of(new String[] { "true", "false" }));
```
ilgrosso: Why not

```java
List.of("true", "false")
```

?

client/idrepo/console/src/main/java/org/apache/syncope/client/console/PreferenceManager.java
```diff
@@ -53,7 +52,7 @@
     private static final TypeReference<Map<String, String>> MAP_TYPE_REF = new TypeReference<Map<String, String>>() {
     };
 
-    private static final List<Integer> PAGINATOR_CHOICES = Arrays.asList(new Integer[] { 10, 25, 50 });
+    private static final List<Integer> PAGINATOR_CHOICES = List.of(new Integer[] { 10, 25, 50 });
```
ilgrosso: Why not

```java
List.of(10, 25, 50)
```

?

client/idrepo/console/src/main/java/org/apache/syncope/client/console/commons/SortableAnyProviderComparator.java
```diff
@@ -33,7 +32,7 @@
 
     private static final long serialVersionUID = 1775967163571699258L;
 
-    private static final Set<String> INLINE_PROPS = new HashSet<>(Arrays.asList(
+    private static final Set<String> INLINE_PROPS = new HashSet<>(List.of(
```
ilgrosso: Why not 
```java
Set.of( "key", "status", "token", "username", "name")
```
?

client/idrepo/console/src/main/java/org/apache/syncope/client/console/pages/Audit.java
```diff
@@ -78,7 +77,7 @@ public Audit(final PageParameters parameters) {
 
             @Override
             protected List<String> getChangeAuthRoles() {
-                return Arrays.asList(
+                return List.of(
```
ilgrosso: Can we avoid `new String[]`?

client/idrepo/console/src/main/java/org/apache/syncope/client/console/panels/AnyTypeClassDetailsPanel.java
```diff
@@ -47,7 +46,7 @@
     private final List<String> availableVirSchemas = schemaRestClient.getVirSchemaNames();
 
     private static final List<String> LAYOUT_PARAMETERS =
-            Arrays.asList(new String[] { "admin.user.layout", "self.user.layout",
+            List.of(new String[] { "admin.user.layout", "self.user.layout",
```
ilgrosso: Same as above

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/content/XMLContentExporter.java
```diff
@@ -90,7 +89,7 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(XMLContentExporter.class);
 
-    private static final Set<String> TABLE_PREFIXES_TO_BE_EXCLUDED = new HashSet<>(Arrays.asList(new String[] {
+    private static final Set<String> TABLE_PREFIXES_TO_BE_EXCLUDED = new HashSet<>(List.of(new String[] {
```
ilgrosso: Same as above

mmoayyed:
This seems to be OK, but we have a failure on appveyor. Is that relevant or related?

ilgrosso:
AppVeyor is more fragile, we keep it mostly to che that build runs on Windows

mmoayyed:
Could we restart the AppVeyor build to see if it might pass again? It's probably, as you note, related to the fragility of the platform and not a real failure.

ilgrosso:
I have no karma to trigger AppVeyor manually, we need a new commit in this PR: just address my latest comments and push a new commit, it should be enough.
Anyway, don't worry too much about AppVeyor...

mmoayyed:
Updated build to address comments, and also converted a bunch more singleton calls to the immutable API calls. Waiting for CI to confirm changes...

mmoayyed:
I might need some help troubleshooting failures here; integration tests are failing and it seems rather difficult to debug and diagnose issues. What's the best approach in debugging an IT?

ilgrosso:
@mmoayyed cd to `fit/core-reference` and start via `mvn -Pdebug,all`, then attach your JPDA debugger to localhost:8000.

You can execute a single test via

```
mvn test -Dtest=RESTITCase#exportInternalStorageContent
```

Finally, please consider that `PullTaskITCase#reconcileFromLDAP` sometimes just fail (the logic behind is quite complex and difficult to synchronize accurately).

mmoayyed:
Thanks. I think I have finally caught and got the hang of this, but this will likely take a few more commits to verify on CI. The cargo plugin seems quite sensitive when it comes to debugging tests, and things hang and ports remain occupied. (It's also likely me, but it does take an unusual amount of time to start the deployment).

ilgrosso:
> The cargo plugin seems quite sensitive when it comes to debugging tests, and things hang and ports remain occupied. (It's also likely me, but it does take an unusual amount of time to start the deployment).

Oh, this sounds strange: I don't have such issues (nor I heard of others), but there might be something with OS X, as you're likely the first one with such OS.

ilgrosso:
Manually merged (due to conflicts) in  6864923b61c395b7be3bcf2e10b5d50aea547515 - thanks @mmoayyed !

Please also note I had to keep https://github.com/apache/syncope/blob/master/fit/core-reference/src/test/java/org/apache/syncope/fit/core/PullTaskITCase.java#L411 because `Map.of()` does not allow for `null` values (and we need exactly that, there).

PR #108, opened 2019-09-04
SYNCOPE-1332: Java 11 language features; use Optional where possible

mmoayyed:
See https://issues.apache.org/jira/browse/SYNCOPE-1332

Convert areas in the codebase where Optional or its variants can be used.


ilgrosso:
Thanks @mmoayyed - honestly I did not check every bit, but globally things look fine - and Travis CI is green, enough to me ;-)

PR #112, opened 2019-09-08
SYNCOPE-1486: upgrade checkstyle; fix syntax issues and line lengths

mmoayyed:
See https://issues.apache.org/jira/browse/SYNCOPE-1486

- Upgrade Checkstyle version
- Re-adjust configuration to locate the right parent for the LineLength check
- Fix instances where line length was violated
- Limit scans to java and xml files to avoid messing up encodings for language bundles.


ilgrosso:
Nice job @mmoayyed !

PR #113, opened 2019-09-09
Added reference to the need for patch in building.

pwd3:
I am trying to do a source build of Syncope on a fresh Centos 7 VM. I still can't get the 
tests to complete (some other problem but have the 

 mvn -PskipTests,all

and I think
 mvn -PskipTests,install

working. To do this I needed to add patch to the list of packages that my Ansible playbook installs on this system. I don't use Windows nor deb package systems much so can't update the site data to include that (nor do I know what will happen on a OSX build).

### DIFF
--- a/src/site/xdoc/building.xml
+++ b/src/site/xdoc/building.xml
@@ -58,6 +58,13 @@ under the License.
         <source>set MAVEN_OPTS=-Xms512m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=512m</source>
       </p>
       <p>
+	You will also need the <a href="https://linux.die.net/man/1/patch">patch</a> program, 
+	on RPM yum style linux systems this can be installed with:
+	<source>sudo yum install patch</source>
+	and on RPM dnf style systems:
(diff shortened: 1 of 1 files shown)


ilgrosso:
@pwd3 not sure why you need to build Syncope sources - normally, you'd do this only if willing to contribute fixes or features.
If you have more questions about this, please send an e-mail to the user@ ML.

Regarding the proposed changes here, please add instructions for Debian or derivatives as follows:

```sudo apt install patch```

In case you are planning to further contribute to Syncope, please fill and send your ICLA: http://syncope.apache.org/contributing#How_do_I_become_a_contributor_or_a_committer.3F

ilgrosso:
Ah, also change your commit message / PR title to include `SYNCOPE-1487` so that JIRA gets automatically linked, thanks.

ilgrosso:
So this PR is for SYNCOPE-1487 and SYNCOPE-1488 ?

ilgrosso:
Reworking this PR as 4b24ca7643b618e734ee8e0cb10287cb29e4ddd5

PR #114, opened 2019-09-09
[SYNCOPE-1490]: incorrect decoding

DmitriyBrashevets:


ilgrosso:
@DmitriyBrashevets as you can see, with your changes there is a failure with `SearchITCase#issueSYNCOPE1223`

Please have a look at https://issues.apache.org/jira/browse/SYNCOPE-1223 and see if your proposed change does make sense, thanks.

DmitriyBrashevets:
> @DmitriyBrashevets as you can see, with your changes there is a failure with `SearchITCase#issueSYNCOPE1223`
> 
> Please have a look at https://issues.apache.org/jira/browse/SYNCOPE-1223 and see if your proposed change does make sense, thanks.

Hi @ilgrosso. Thanks for your reply. I see now..
Does it mean that I first need to encode filter property values (to support commas) and after this, I need to encode full fiql string?

Code sample:
```java
String fiqlString = SyncopeClient.getUserSearchConditionBuilder()
                .is("phone").equalTo(utf8Encode("+1234556789112"))
                .query();
AnyQuery anyQuery = new AnyQuery.Builder().realm(SyncopeConstants.ROOT_REALM).fiql(utf8Encode(fiqlString)).build();
        PagedResult<UserTO> matchingUsers = userService.search(anyQuery);
```

If so, I'll close the PR and corresponding bug that I created in JIRA.

ilgrosso:
> Does it mean that I first need to encode filter property values (to support commas) and after this, I need to encode full fiql string?

Yes, please look at my comment in JIRA.

PR #115, opened 2019-09-11
SYNCOPE-1489: Reorganize travis CI into build stages/jobs

mmoayyed:
This is the initial step at changing travis CI to run under multiple stages and jobs. This *first pass* creates a very basic build that also verifies Javadocs and Checkstyle, etc. Follow PRs will modify the CI to include tests and FITs, etc. As part of this PR, issues related to Javadoc failures are fixed that affect improper tag usage.

There is a small thing with the checkstyle job that needs to be worked out, but this is good enough for now and for reviews.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,18 +17,52 @@ sudo: false
 language: java
 jdk:
   - openjdk11
-before_install:
-  - echo 'MAVEN_OPTS="-Xmx4096M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+TieredCompilation -XX:TieredStopAtLevel=1"' >~/.mavenrc
-# default install is mvn install --quiet -DskipTests=true
-install: travis_wait 30 mvn --show-version --quiet -T 1C -P all,skipTests
-before_script:
-  - cd fit/core-reference
-script:
-  # invoker.streamLogs: we cannot access to log files through Travis web ui, so display everything in the console
-  - mvn --show-version verify -Dinvoker.streamLogs=true
-after_failure:
-  - cat target/log/*
-  - cat target/failsafe-reports/org.apache.syncope.fit.*-output.txt
+cache:
+bundler: false
+cargo: false
+directories:
+  - $HOME/.m2
+git:
+  depth: 2
+matrix:
+  fast_finish: true
+env:
+  global:
+  - MAVEN_OPTS="-Xmx4096M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+TieredCompilation -XX:TieredStopAtLevel=1"
+install: true
 notifications:
   email:
     - dev@syncope.apache.org
+after_failure:
+  - cat target/log/*
+  - cat target/failsafe-reports/org.apache.syncope.fit.*-output.txt
+stages:
+  - initialize
+  - build
+  - validate
+  - test
+  - finalize
+jobs:
+  include:
+    - stage: initialize
+      script: mvn --show-version -T 4 clean dependency:go-offline dependency:resolve-plugins
+      name: "Initialize Build System"
+      #######################################################
+    - stage: build
+      script: mvn install --show-version --quiet -T 4 -P all,skipTests -Dcheckstyle.skip=true -Dmaven.javadoc.skip=true
+      name: "Build"
+      #######################################################
+    - stage: validate
+      script: mvn checkstyle:check --show-version -T 4
+      name: "Checkstyle Analysis"
+    - stage: validate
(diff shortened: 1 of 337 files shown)


ilgrosso:
@mmoayyed thanks for this PR and also for fixing Javadocs!

src/site/xdoc/building.xml
```diff
@@ -85,7 +85,7 @@ under the License.
       <p>
         To build Syncope documentation execute (from within the top-level source directory):
         <source>$ mvn -N -P site clean generate-resources</source>
-        The documentation artifacts are now be available under <tt>target/generated-docs/</tt>.
+        The documentation artifacts are now be available under <code>target/generated-docs/</code>.
```
ilgrosso: Are the changes under `src/site/xdoc` required? Not sure if they are going to badly affect the project website generation...
If ensure, please revert, thanks.

.travis.yml
```diff
@@ -17,18 +17,65 @@ sudo: false
...
+      name: "Javadoc Analysis"
+    - stage: validate
+      script: mvn -T 4 --quiet ianal:verify-legal-files -Denforcer.skip=true
+      name: "Legal Files Analysis"
+    - stage: validate
+      script: mvn -T 4 --quiet modernizer:modernizer
+      name: "Modernizer Legacy API Analysis"
+      ######################################################
+    - stage: test
+      script: mvn -f fit/core-reference/pom.xml verify --quiet -T 4 -Dinvoker.streamLogs=true -Denforcer.skip=true -Dmodernizer.skip=true -Djacoco.skip=true
```
ilgrosso: I have seen the usual error in https://travis-ci.org/apache/syncope/jobs/583783048

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```

Maybe you can change this line to

````
travis_wait 30 mvn -f fit/core-reference/pom.xml verify -T 4 -Dinvoker.streamLogs=true -Denforcer.skip=true -Dmodernizer.skip=true -Djacoco.skip=true
```

e.g. remove `--quiet` and raise the silence timeout to 30 minutes instead of 10.

mmoayyed: This will likely do the job, but does have the drawback that all build output is basically silenced (to whatever degree we have them on) because the job is basically run in the background.

I'll wait to see if the build passes, but this strategy likely needs to be reformulated to be something like this instead:

```
waitloop="while sleep 9m; do echo -e '\n=====[build is still running ]====='; done &"
eval $waitloop
waitRetVal=$?
```

And then we execute the maven task. This is basically the same thing as `travis_wait` except output is kept alive from the build.

mmoayyed:
This LGTM as far as the changeset goes, but we do have a few failures:

```
[ERROR] Failures: 
[ERROR]   PullTaskITCase.reconcileFromLDAP:444 expected: <null> but was: <org.apache.syncope.common.lib.Attr@f36afeef>
[ERROR]   PushTaskITCase.pushMatchingUser:238 expected: <false> but was: <true>
```

The usual suspect being `PullTaskITCase.reconcileFromLDAP`, but I'm not sure if the other failure is related here.

ilgrosso:
I restarted the job to see if this is permanent - as you can see, no such errors in https://travis-ci.org/apache/syncope/builds/583713228 (latest `master` build).

ilgrosso:
It seems last run failed at last stage: https://travis-ci.org/apache/syncope/jobs/584042105

and the reason was

```
The job exceeded the maximum log length, and has been terminated.
```

Moreover:
1. do you think my comment above, about `modernizer-maven-plugin` can be easily addressed?
1. can we add a stage before IT which runs unit tests, e.g. something like as `mvn clean package` from root?

mmoayyed:
1. Modernizer should already be included in previous commits. Please see https://travis-ci.org/apache/syncope/builds/584042094?utm_source=github_status&utm_medium=notification in the `Validate` phase. 
2. Sure. I'll add that too.

mmoayyed:
Closing PR for now to minimize notifications. Will re-open again once the changeset is ready again.

mmoayyed:
I had to take out the versions plugin; it's quite inefficient it seems and takes a very long time to complete. Dependency updates (and reports at least) can be automated in different and better ways in the future.

ilgrosso:
Hey @mmoayyed Travis is finally green (and AppVeyour is red but due to the usual instability); since your changes look very good to me, shall I just go ahead and merge?!?

PR #116, opened 2019-09-17
NOJIRA - Small improvements/changes from static analysis reports

mmoayyed:
This pull request is quite large, but it is also largely innocent. It presents the following:

- Remove toString() calls on arrays
- Make sure collected collections as return values are used
- Replace String literals of length one being used in concatenation, gaining some performance enhancement.
- Mark relevant methods (private or final) as static where appropriate, gaining small performance enhancement.
- Add RAT checks to CI independently, removing them from the verify phase for FITs.


.travis.yml
```diff
@@ -64,13 +64,16 @@ jobs:
...
+      script: mvn -T 4 --quiet apache-rat:check
+      name: "Release Audit Tool (RAT) Analysis"
       ######################################################
     - stage: test
       script: mvn clean test -T 4 -P skipLegalFiles,skipSass -T 4 --quiet -Dinvoker.streamLogs=true -Denforcer.skip=true -Dmodernizer.skip=true
       name: "Unit Tests"
       ######################################################
     - stage: fit
-      script: travis_wait 30 mvn -f fit/core-reference/pom.xml verify -P skipLegalFiles,skipSass -T 4 -Dinvoker.streamLogs=true -Denforcer.skip=true -Dmodernizer.skip=true -Djacoco.skip=true
+      script: travis_wait 45 mvn -f fit/core-reference/pom.xml verify -P skipLegalFiles,skipSass -T 4 -Dinvoker.streamLogs=true -Denforcer.skip=true -Dmodernizer.skip=true -Djacoco.skip=true -Drat.skip=true
```
ilgrosso: here you can add `-P full-it,all` to trigger Admin Console and Enduser UI integration tests too

mmoayyed: Can those tests be run independently in a separate job perhaps to save time, with another profile perhaps? 

Or maybe we can add two more jobs with these?

- travis_wait 45 mvn -f fit/enduser-reference/pom.xml ...
- travis_wait 45 mvn -f fit/console-reference/pom.xml ...

..giving each 45 minutes. That might work, time-wise?

ilgrosso: The tests are still in `fit/core-reference`, but excluded by default.

So it should be something like as:
```yaml
script: travis_wait 45 mvn -f fit/core-reference/pom.xml verify -P full-it,all,skipLegalFiles,skipSass -T 4 -Dinvoker.streamLogs=true -Denforcer.skip=true -Dmodernizer.skip=true -Djacoco.skip=true -Drat.skip=true
```

mmoayyed: Got it, will do.

I see some IT tests inside the enduser-reference module for OIDC and SAML2SP. Should we enable and run those in the CI? Not quite sure how they are executed now.

ilgrosso: > I see some IT tests inside the enduser-reference module for OIDC and SAML2SP. Should we enable and run those in the CI? Not quite sure how they are executed now.

They require Docker, guess we'll handle them later, no?

mmoayyed: Absolutely. 

I guess [this page](https://syncope.apache.org/building#fit.2Fconsole-reference) should be slightly updated to make the docker requirement clearer. Today's reading suggests that docker is only required for DB tests and console/enduser tests can just run on their own...or am I reading that wrong?

ilgrosso: That page mentions `fit/enduser-reference` and `fit/console-reference` only for `debug` and `hotswap` profiles, no tests.

Moreover, the profile `sso-it` in `fit/enduser-reference` is only available in the `master` branch at the moment.

ilgrosso:
I was wondering: if all methods from *RestClient classes are now static, why are such classes still instantiated?

I would then also remove all statements like as https://github.com/apache/syncope/blob/master/client/idrepo/console/src/main/java/org/apache/syncope/client/console/status/ChangePasswordModal.java#L47 and turn all instance invocations into static invocations.

.travis.yml
mmoayyed: I'll take note in the follow-up JIRA when I rework ITs based on Docker. Thank you!

mmoayyed:
> I was wondering: if all methods from *RestClient classes are now static, why are such classes still instantiated?

Sure thing; I'll re-adjust. I originally positioned the analysis to only affect candidate methods marked as private to avoid compile issues; now after this pass, we can certainly clean up some more :)

mmoayyed:
I added two additional tests in the verification phase to test content-type values for YAML, etc. Looks like tests are failing with NPEs. Are these still valid tests and should I look into them?

ilgrosso:
Well, in principle they should just work but in practice I don't think that YAML or XML are much used...

I'd propose to remove (comment out?) the last two stages from Travis CI for the moment, then open two issues on JIRA for the test issues; when such issues are resolved we can also re-enable the stages.

WDYT?

mmoayyed:
Sure, I took them out for now.

As for the remaining changes, I managed to clean up quite a bit. There are still a few areas where classes with all static methods are initialized, specially in *Panel components I think and I'll try to handle them separately in subsequent pull requests. This should be ready to merge as far as I can tell.

ilgrosso:
Thanks again for your work @mmoayyed !

PR #118, opened 2019-09-23
NOJIRA: Clean up SpringBoot set up and @Bean methods

mmoayyed:
This pull request does a modest amount of cleaning up Spring Boot configuration and/or bean methods

- Marks a number of bean methods as conditional so they can be overridden by deployers without overlaying the whole class, etc.
- Merges relevant Spring Boot annotations together
- Allows bean methods to be named by their respective method names to keep things consistent across the codebase.


core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/MasterDomain.java
```diff
@@ -149,34 +156,37 @@ public DomainEntityManagerFactoryBean masterEntityManagerFactory() {
         DomainEntityManagerFactoryBean masterEntityManagerFactory = new DomainEntityManagerFactoryBean();
         masterEntityManagerFactory.setMappingResources(orm);
         masterEntityManagerFactory.setPersistenceUnitName("Master");
-        masterEntityManagerFactory.setDataSource((DataSource) masterDataSource().getObject());
+
```
ilgrosso: I don't get the rationale of this change, e.g. from

```
 masterEntityManagerFactory.setDataSource((DataSource) masterDataSource().getObject());
```

to

```
        DataSource dataSource = (DataSource) masterDataSource().getObject();
        masterEntityManagerFactory.setDataSource(Objects.requireNonNull(dataSource));
```

Why not leaving it as single statement?

mmoayyed: Static analysis did complain that `(DataSource) masterDataSource().getObject();` can in in fact be null, which would then ultimately cause an NPE down the road. It appears that underlying components do catch null at various levels, but it might be good to prevent it upfront.

ilgrosso: Agree, but why not simply changing into:

```
 masterEntityManagerFactory.setDataSource(Objects.requireNonNull((DataSource) masterDataSource().getObject()));
```

mmoayyed: Sure thing.

ilgrosso:
@mmoayyed does it work for you locally? Travis CI keeps failing: https://travis-ci.org/apache/syncope/jobs/588385015

ilgrosso:
Thanks @mmoayyed Travis CI is now green :+1:

PR #120, opened 2019-09-23
NOJIRA: Protect invocations against NPEs and illegal sort ops

mmoayyed:
This pull request is a semi-large cleanup effort to protect against the following issues:

- Immutable object modification as part of sort operations.
- Null check clean up where the condition is always true/false
- Protect against NPEs where objects may be calculated as `null`
- Switch sort ops to use factory methods such as `List.sort()` instead of `Collections.sort()`

All FITs should pass, as they did locally.

### DIFF
--- a/client/idrepo/enduser/src/main/java/org/apache/syncope/client/enduser/pages/SelfConfirmPasswordReset.java
+++ b/client/idrepo/enduser/src/main/java/org/apache/syncope/client/enduser/pages/SelfConfirmPasswordReset.java
@@ -69,7 +69,7 @@ public class SelfConfirmPasswordReset extends BaseEnduserWebPage {
     public SelfConfirmPasswordReset(final PageParameters parameters) {
         super(parameters);
 
-        if (parameters == null || parameters.get("token").isEmpty()) {
+        if (parameters != null || parameters.get("token").isEmpty()) {
             LOG.debug("No token parameter found in the request url");
             parameters.add("errorMessage", getString("self.confirm.pwd.reset.error.empty"));
             setResponsePage(getApplication().getHomePage(), parameters);
--- a/core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
+++ b/core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
@@ -275,7 +275,7 @@ public class SearchCond extends AbstractSearchCond {
         String anyTypeName = null;
 
         if (type == null) {
-            return anyTypeName;
+            return null;
         }
 
         switch (type) {
@@ -363,7 +363,7 @@ public class SearchCond extends AbstractSearchCond {
         boolean isValid = false;
 
         if (type == null) {
-            return isValid;
+            return false;
         }
 
         switch (type) {
--- a/client/am/console/src/main/java/org/apache/syncope/client/console/panels/GatewayRouteDirectoryPanel.java
+++ b/client/am/console/src/main/java/org/apache/syncope/client/console/panels/GatewayRouteDirectoryPanel.java
@@ -21,7 +21,6 @@ package org.apache.syncope.client.console.panels;
 import de.agilecoders.wicket.core.markup.html.bootstrap.dialog.Modal;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import org.apache.commons.lang3.SerializationUtils;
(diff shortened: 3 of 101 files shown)


client/idrepo/enduser/src/main/java/org/apache/syncope/client/enduser/pages/SelfConfirmPasswordReset.java
```diff
@@ -69,7 +69,7 @@
     public SelfConfirmPasswordReset(final PageParameters parameters) {
         super(parameters);
 
-        if (parameters == null || parameters.get("token").isEmpty()) {
+        if (parameters != null || parameters.get("token").isEmpty()) {
```
ilgrosso: Why changing this check? Please revert.

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
```diff
@@ -275,7 +275,7 @@ public String hasAnyTypeCond() {
         String anyTypeName = null;
 
         if (type == null) {
-            return anyTypeName;
+            return null;
```
ilgrosso: What's the rationale behind such change?

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
```diff
@@ -363,7 +363,7 @@ public boolean isValid() {
         boolean isValid = false;
 
         if (type == null) {
-            return isValid;
+            return false;
```
ilgrosso: What's the rationale behind such change?

client/idrepo/enduser/src/main/java/org/apache/syncope/client/enduser/pages/SelfConfirmPasswordReset.java
mmoayyed: Original:

```java
        if (parameters == null || parameters.get("token").isEmpty()) {
            LOG.debug("No token parameter found in the request url");
            parameters.add("errorMessage", getString("self.confirm.pwd.reset.error.empty")); // No good. NPE if parameters is `null`
            setResponsePage(getApplication().getHomePage(), parameters);
        }
```

See https://github.com/apache/syncope/blob/master/client/idrepo/enduser/src/main/java/org/apache/syncope/client/enduser/pages/SelfConfirmPasswordReset.java#L72

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
mmoayyed: Original variable name is always `null` here. It's best to be explicit about what is returned, rather than letting the compiler dereference later. Makes it difficult to read.

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
mmoayyed: Same as above. Original variable name is always `false`

client/idrepo/enduser/src/main/java/org/apache/syncope/client/enduser/pages/SelfConfirmPasswordReset.java
ilgrosso: The original check is like "if parameters is null, or, if not, if there is a token"; your change is like "if parameters is not null and there is a token".

I don't understand why you are  proposing to change the check.

mmoayyed: If the parameters is null, the condition and the flow continues on where you have a chance of NPE. The change attempts to fix that, but you're right in that I am not quite sure what the behavior should be here. Clearly, if the `parameters` is null, we can't add anything to it, but what should happen if it's null anyway? can it be null at all?

mmoayyed: Perhaps:

```
if (parameters != null && parameters.get("token").isEmpty()) {
}
```

Better?

ilgrosso: I was missing the `parameters.add()` inside the if block; at this point your latest proposal is the best one, I agree.

ilgrosso: Sorry, I was meaning: let's go ahead with your original proposal, let's resolve this.

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
ilgrosso: Hmm,  not sure about this, but also no strong opinion.
Let's leave it as per your change.

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/search/SearchCond.java
ilgrosso: See above.

PR #122, opened 2019-09-24
NOJIRA: Enable Apache Maven wrapper plugin

mmoayyed:
This pull request enables the [maven-wrapper for Syncope](https://github.com/takari/maven-wrapper). Main objectives are:

1. Local builds do not need to have (the right version of ) Apache Maven locally installed; specially relevant for Windows.
2. Control the build via the proper Apache Maven version. 

The command to update the wrapper would be: `mvn -N io.takari:maven:0.7.6:wrapper` (or whatever is the version of the plugin)

Notes:

> By default, the Maven Wrapper JAR archive is added to the using project as small binary file .mvn/wrapper/maven-wrapper.jar. It is used to bootstrap the download and invocation of Maven from the wrapper shell scripts. If your project is not allowed to contain binary files like this, you can configure your version control system to exclude checkin/commit of the wrapper jar.

> If the JAR is not found to be available by the scripts they will attempt to download the file from the URL specified in .mvn/wrapper/maven-wrapper.properties under wrapperUrl and put it in place. The download is attempted via curl, wget and, as last resort, by compiling the ./mvn/wrapper/MavenWrapperDownloader.java file and executing the resulting class.


ilgrosso:
@mmoayyed I am not sure of the benefits of adding this: would you mind initiating a `[DISCUSS]` thread on dev@ on the topic? Thanks!

mmoayyed:
Based on discussions on the dev mailing list, we're going to punt this one.

ilgrosso:
FTR: https://lists.apache.org/thread.html/1b782cbac24508fd7de96aabafd9ee198e8f3785a446bb016ef8fbd7@%3Cdev.syncope.apache.org%3E

PR #124, opened 2019-10-01
SYNCOPE-1465: Add executor information to Report executions

mmoayyed:
- [JIRA]( https://issues.apache.org/jira/browse/SYNCOPE-1465): adds an `executor` field to `ReportTO` objects, etc to indicate the authenticated username that starts the report.
- Follow-up work with be done in another pull request to handle tasks in similar ways.
- Also fixes a typo in `addFurtherAcions` (missing `t`).

Screenshot:

image


core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/job/report/DefaultReportJobDelegate.java
```diff
@@ -124,6 +127,8 @@ public void execute(final String reportKey) throws JobExecutionException {
         execution.setStatus(ReportExecStatus.STARTED);
         execution.setStart(new Date());
         execution.setReport(report);
+
+        execution.setExecutor(getAuthenticationPrincipal());
```
ilgrosso: Replace this with `org.apache.syncope.core.spring.security.AuthContextUtils#getUsername()`

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/job/report/DefaultReportJobDelegate.java
```diff
@@ -223,4 +228,12 @@ public void execute(final String reportKey) throws JobExecutionException {
             reportExecDAO.save(execution);
         }
     }
+
```
ilgrosso: This method can be safely removed

client/idrepo/console/src/main/java/org/apache/syncope/client/console/reports/ReportDirectoryPanel.java
```diff
@@ -108,7 +108,7 @@ protected void onTimer(final AjaxRequestTarget target) {
 
         columns.add(new KeyPropertyColumn<>(new StringResourceModel("key", this), "key"));
         columns.add(new PropertyColumn<>(new StringResourceModel("name", this), "name", "name"));
-
+
```
ilgrosso: formatting-only change?

client/idrepo/console/src/main/resources/org/apache/syncope/client/console/tasks/ExecutionsDirectoryPanel.properties
```diff
@@ -17,4 +17,5 @@
 start=Start
 end=End
 status=Status
+executor=Executor
```
ilgrosso: Please also add the new label to other languages' `ExecutionsDirectoryPanel_*.properties`, even with english value.
It's better to keep them all in sync.

mmoayyed: What would be the right translation for Italian?

ilgrosso: Esecutore

ilgrosso:
It seems Travis CI is failing: https://travis-ci.org/apache/syncope/jobs/591982376#L725-L727

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/AbstractExec.java
```diff
@@ -38,6 +38,8 @@
     @NotNull
     protected String status;
 
+    protected String executor;
```
ilgrosso: Why there is no `@NotNull` here?

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/job/report/DefaultReportJobDelegate.java
```diff
@@ -124,6 +126,8 @@ public void execute(final String reportKey) throws JobExecutionException {
         execution.setStatus(ReportExecStatus.STARTED);
         execution.setStart(new Date());
         execution.setReport(report);
+
+        execution.setExecutor(AuthContextUtils.getUsername());
```
ilgrosso: Report execution can be either triggered [via REST](https://github.com/apache/syncope/blob/master/core/idrepo/logic/src/main/java/org/apache/syncope/core/logic/ReportLogic.java#L163) or scheduled; either way, it will be submitted to Quartz.

[ReportJob](https://github.com/apache/syncope/blob/master/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/job/report/ReportJob.java) is executed by Quartz, and in the block https://github.com/apache/syncope/blob/master/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/job/report/ReportJob.java#L64-L74 will invoke the configured `ReportJobDelegate` instance, which is by default this class, e.g. `DefaultReportJobDelegate`.

As you can see, an instrumented authentication context is built there for user `admin`; as a result, `AuthContextUtils.getUsername()` will always return `admin` here, which, at least for executions triggered via REST, may not be correct.

In order to get the actual executor, you should invoke `AuthContextUtils.getUsername()` in `ReportLogic#execute`, then pass such value to `jobManager#register` and add it in the Map that will be passed inside `JobExecutionContext` to `ReportJob#execute` an replace there `AuthContextUtils#callAsAdmin` with `AuthContextUtils#callAs`.

Naturally, you should also adjust things so that `admin` is used when execution trigger is not coming via REST but scheduled.

ilgrosso: Forgot to add: it would be more or less the same for Tasks.

mmoayyed: Thanks very much for the notes; the design makes much better sense now. Will adjust.

ilgrosso: After thinking again to the change I propose above, I believe we should keep calling `AuthContextUtils#callAsAdmin` and add `executor` as String param to `ReportJobDelegate#execute` - it's simpler and also more in line with similar use cases as audit and notifications.

mmoayyed: Question:

If I modify the `JobManagerImpl` to be:

```
public Map<String, Object> register(final SchedTask task, final Date startAt, 
									final long interruptMaxRetries,
                                    final String executor) {

	...
	jobMap.put(JobManager.EXECUTOR_KEY, executor);
    ...
}
```


That would mean that I also may need to modify:

```
public void load(final String domain, final DataSource datasource) {}
```

Because, it calls the `register` function [here](https://github.com/apache/syncope/blob/master/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/job/JobManagerImpl.java#L353) in `load()`, and this would cause API changes to many other components.

So, in `JobManagerImpl#load()`, we could just call `AuthContextUtils.getUsername()` such that

```
register(task, task.getStartAt(), conf.getRight(), AuthContextUtils.getUsername());
```

Would that work?

ilgrosso: > Would that work?

Nope. 

`JobManagerImpl#load()` is invoked at startup: at that point, no authentication context exists, of course.

Hence, you can add

```java
    @Resource(name = "adminUser")
    private String adminUser;
```

to `JobManagerImpl` and then invoke

```java
register(task, task.getStartAt(), conf.getRight(), adminUser);
```

Incidentally, this is exactly what I mean above by saying that report and task execution can be scheduled.

fit/core-reference/src/test/java/org/apache/syncope/fit/core/ReportTemplateITCase.java
```diff
@@ -36,11 +36,24 @@
...
 import org.apache.syncope.fit.AbstractITCase;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.userdetails.User;
 
 public class ReportTemplateITCase extends AbstractITCase {
 
+    @BeforeAll
```
ilgrosso: What is this `setup()` method supposed to do? I think you could safely drop it.

fit/core-reference/src/test/java/org/apache/syncope/fit/core/ReportITCase.java
```diff
@@ -54,12 +54,25 @@
...
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.userdetails.User;
 
 public class ReportITCase extends AbstractITCase {
 
+    @BeforeAll
+    public static void setup() {
```
ilgrosso: What is this `setup()` method supposed to do? I think you could safely drop it.

ilgrosso:
@mmoayyed it seems that [Travis CI](https://travis-ci.org/apache/syncope/jobs/592984632) keeps failing by timeout in integration tests (I manually restarted that job three times, always same result).
It also seems that [AppVeyor](https://ci.appveyor.com/project/ApacheSoftwareFoundation/syncope/builds/27850310) shows similar troubles: can you figure out why? I think your local build is successful, no?

mmoayyed:
Hmmm strange. I'll run this one more time locally but yes mine seems to pass. It's possible CI is somehow running out of memory and hanging but I can't imagine that would be because of the changes here. I'll report back shortly.

mmoayyed:
Update: Had to invalidate caching to let maven resume ITs and now I see the failures. Most are failing due to executor fields being `null` for various task types, or areas where executor is not calculated or set. Been working to make further progress and fix test cases, and there is still quite a bit of work to make sure all ITs can pass.

mmoayyed:
Cleaning up tests; I have had to simulate an authentication context for ReportITCase and ReportTemplateITCase specially to make sure executions have a valid executor now that the field is marked as non-null. Still testing.

mmoayyed:
Waiting for CI to pass; I think in theory, this patch does also handle accounting for executors for tasks as well. I will nonetheless take a separate pass to make sure and follow-up with another PR if necessary.

mmoayyed:
Last failure was quite strange:

```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:31 min (Wall Clock)
[INFO] Finished at: 2019-10-07T13:07:37Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:testCompile (default-testCompile) on project syncope-fit-core-reference: Compilation failure
[ERROR] /home/travis/build/apache/syncope/fit/core-reference/src/test/java/org/apache/syncope/fit/core/ReportITCase.java:[99,35] cannot find symbol
[ERROR]   symbol:   method getLastExecutor()
[ERROR]   location: variable reportTO of type org.apache.syncope.common.lib.to.ReportTO
```

I wonder if this is due to caching on Travis CI. Started the CI again with an empty commit to make sure...

PR #126, opened 2019-10-09
[SYNCOPE-1502]: find anytos using FIQL

DmitriyBrashevets:
PR contains improvements when AnyTOs are searched using FIQL queries:

- AnyTO objects are searched by keys retrieved from the search view using a single SQL query that contains IN condition
-  replacement of multiple OR-realm condition in the resulted SQL query that is executed to search entities in search views with IN clause.

### DIFF
--- a/core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/AnyDAO.java
+++ b/core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/AnyDAO.java
@@ -37,6 +37,8 @@ public interface AnyDAO<A extends Any<?>> extends DAO<A> {
 
     String findKey(String name);
 
+    List<A> findByKeys(List<String> keys);
+
     Date findLastChange(String key);
 
     A authFind(String key);
--- a/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/AbstractAnySearchDAO.java
+++ b/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/AbstractAnySearchDAO.java
@@ -21,8 +21,12 @@ package org.apache.syncope.core.persistence.jpa.dao;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.validation.ValidationException;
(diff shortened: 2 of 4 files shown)


core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/dao/AnyDAO.java
```diff
@@ -37,6 +39,8 @@
 
     String findKey(String name);
 
+    List<A> findByKeys(Set<String> keys);
```
ilgrosso: Would you mind changing this method's signature into

```java
List<A> findByKeys(List<String> keys);
```

This will ensure we keep same ordering between input and output.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/AbstractAnySearchDAO.java
```diff
@@ -297,26 +301,37 @@ protected String check(final MemberCond cond) {
     }
 
     protected <T extends Any<?>> List<T> buildResult(final List<Object> raw, final AnyTypeKind kind) {
-        List<T> result = new ArrayList<>();
+        LinkedHashSet<String> orderedAnyKeys = extractSortedAnyKeys(raw);
```
ilgrosso: Any reason not to use `List` instead? e.g.

```java
List<String> orderedAnyKeys = extractSortedAnyKeys(raw);
```

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/AbstractAnySearchDAO.java
```diff
@@ -297,26 +301,37 @@ protected String check(final MemberCond cond) {
...
-                });
-
-        return result;
+                : ((String) anyKey))
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T extends Any<?>> List<T> findAnyTOs(AnyTypeKind kind, Set<String> orderedAnyKeys) {
+        if (AnyTypeKind.USER == kind) {
```
ilgrosso: Please remove this method and use instead

```java
anyUtilsFactory.getInstance(kind).dao().findByKeys(orderedAnyKeys);
```

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/AbstractAnySearchDAO.java
DmitriyBrashevets: Ok, as SQL returns distinct keys, I'll return list.

ilgrosso:
@DmitriyBrashevets Travis CI is failing: https://travis-ci.org/apache/syncope/builds/595621147?utm_source=github_status&utm_medium=notification

Please check and fix, thanks.

ilgrosso:
@DmitriyBrashevets please also:
1. create an improvement on JIRA for this work
1. change your commit message to include the full JIRA number (e.g. `[SYNCOPE-150X]`)

thanks.

DmitriyBrashevets:
> @DmitriyBrashevets please also:
> 
> 1. create an improvement on JIRA for this work
> 2. change your commit message to include the full JIRA number (e.g. `[SYNCOPE-150X]`)
> 
> thanks.

Done

ilgrosso:
Thanks for creating SYNCOPE-1502.

ilgrosso:
@DmitriyBrashevets I am making some adjustments, expect a merge soon

ilgrosso:
I added a [second commit](68cfc0d44b79539a28e36b5322d444a9b1a7d7cc) - to keep yours untouched - where:

1. I fixed few checkstyle violations (you'll have to build from root folder via `mvn clean install` to perform all necessary checks that Travis CI mostly skips, due to timing constraints)
1. I removed one-line, single-use methods
1. I verified, by running the whole integration test suite against the different environments via the various IT profiles under `fit/core-reference/pom.xml` - the compatibility of your changes with:
    * PostgreSQL
    * MySQL
    * PostgreSQL JSONB
    * MySQL JSON (which for some reason did not like the new `getAdminRealmsFilter()`, so I had to keep it old-style just for it)
    * Elasticsearch

All works fine, thanks @DmitriyBrashevets for your contribution.

ilgrosso:
@DmitriyBrashevets FYI 6bd9cf3e6106d1f755b3beb7915fb2dfdcc90c1c

PR #127, opened 2019-10-10
SYNCOPE-1465: make sure task objects track the executor.

mmoayyed:
Also make sure FITs have test cases that enforces this.

### DIFF
--- a/core/idrepo/logic/src/main/java/org/apache/syncope/core/logic/NotificationLogic.java
+++ b/core/idrepo/logic/src/main/java/org/apache/syncope/core/logic/NotificationLogic.java
@@ -70,7 +70,8 @@ public class NotificationLogic extends AbstractJobLogic<NotificationTO> {
 
     @PreAuthorize("hasRole('" + IdRepoEntitlement.NOTIFICATION_CREATE + "')")
     public NotificationTO create(final NotificationTO notificationTO) {
-        return binder.getNotificationTO(notificationDAO.save(binder.create(notificationTO)));
+        Notification notification = notificationDAO.save(binder.create(notificationTO));
+        return binder.getNotificationTO(notification);
     }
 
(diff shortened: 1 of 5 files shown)


core/idrepo/logic/src/main/java/org/apache/syncope/core/logic/NotificationLogic.java
```diff
@@ -70,7 +70,8 @@ public NotificationTO read(final String key) {
 
     @PreAuthorize("hasRole('" + IdRepoEntitlement.NOTIFICATION_CREATE + "')")
     public NotificationTO create(final NotificationTO notificationTO) {
-        return binder.getNotificationTO(notificationDAO.save(binder.create(notificationTO)));
+        Notification notification = notificationDAO.save(binder.create(notificationTO));
```
ilgrosso: What's the rationale for this change?

mmoayyed: Chaining is good, but I'd say it's easier to read and understand this way.

ilgrosso: Hum, I don't agree completely: anyway, I still don't see the rationale of such change in the context of the current PR.

If you really want to propose such change, you should be doing it across a whole lot of Logic classes' methods, which all look as this one.

mmoayyed: You're right. I like things to be consistent all the way. So let me revert this.

ilgrosso:
Build https://travis-ci.org/apache/syncope/builds/595974719 is fine, and the only afterward change is ad10777071307042e448838d3740591f6c83f461 - no need to wait, let's merge.

PR #133, opened 2019-10-17
[SYNCOPE-1503] Provision rules can now be deleted from Resource

FedericoPalmitesta:


ilgrosso:
Thanks @FedericoPalmitesta for spotting this!

PR #141, opened 2019-11-28
SYNCOPE-1511: History management for admin console UI

mmoayyed:
This pull request adds the ability to review history for users, groups and any-objects. The history that is displayed is narrowed down to show create/update events where the result is always success.


client/console/src/main/java/org/apache/syncope/client/console/audit/AuditEntryBean.java
```diff
@@ -0,0 +1,200 @@
...
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.syncope.common.lib.to.AnyTO;
+
+import java.io.Serializable;
+import java.util.Date;
+import java.util.List;
+
+public class AuditEntryBean implements Serializable {
```
ilgrosso: I'd rename it `AnyTOAuditEntryBean` to highlight the fact that id wraps an `AnyTO` - we are likely to introduce `ConnInstanceTOAuditEntryBean` and `ResourceTOAuditEntryBean` in the near future, no?

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,187 @@
...
+        final AnyTO anyTO) {
+
+        super(MultilevelPanel.FIRST_LEVEL_ID, pageRef);
+        disableCheckBoxes();
+        
+        this.baseModal = baseModal;
+        this.multiLevelPanelRef = multiLevelPanelRef;
+        this.anyTO = anyTO;
+
+        if (anyTO instanceof UserTO) {
```
ilgrosso: Replace this chain with
```
anyTypeKind = AnyTypeKind.fromTOClass(anyTO.getClass());
```

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    protected String paginatorRowsKey() {
+        return Constants.PREF_AUDIT_HISTORY_PAGINATOR_ROWS;
+    }
+
+    @Override
+    protected List<IColumn<AuditEntryBean, String>> getColumns() {
+        final List<IColumn<AuditEntryBean, String>> columns = new ArrayList<>();
+        columns.add(new KeyPropertyColumn<>(
+            new StringResourceModel("key", this), "key"));
```
ilgrosso: There is no need of a `key` column: all rows will always have the same value, matching the user / group / any object selected.

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,187 @@
...
+            return getAuditEntryBeans(first, count).iterator();
+        }
+
+        @Override
+        public long size() {
+            return restClient.count(anyTO.getKey(), getQueryableAuditEvents(), getQueryableAuditResults());
+        }
+
+        @Override
+        public IModel<AuditEntryBean> model(final AuditEntryBean auditEntryBean) {
```
ilgrosso: Is this method override really needed?

client/console/src/main/java/org/apache/syncope/client/console/rest/AuditHistoryRestClient.java
```diff
@@ -0,0 +1,69 @@
...
+            .key(key)
+            .page(page)
+            .events(events)
+            .results(results)
+            .orderBy(toOrderBy(sort))
+            .build();
+        return getService(AuditService.class).search(query).getResult();
+    }
+
+    public int count(final String key) {
```
ilgrosso: Why two `count()` methods?

client/console/src/main/java/org/apache/syncope/client/console/rest/AuditHistoryRestClient.java
```diff
@@ -0,0 +1,70 @@
...
+            .size(size)
+            .page(page)
+            .events(events)
+            .result(result)
+            .orderBy(toOrderBy(sort))
+            .build();
+        return getService(AuditService.class).search(query).getResult();
+    }
+
+    public List<AuditEntryTO> search(final String key,
```
ilgrosso: Why is this non-paging method required?

common/lib/src/main/java/org/apache/syncope/common/lib/to/AuditEntryTO.java
```diff
@@ -39,11 +39,11 @@
 
     private String result;
 
-    private String before;
+    private Object before;
```
ilgrosso: Why String -> Object?

client/console/src/main/java/org/apache/syncope/client/console/rest/AuditHistoryRestClient.java
mmoayyed: I think this is used to populate the drop-down on the modal dialog with history items to compare; a similar approach seems to be [used here](https://github.com/apache/syncope/blob/2_1_X/client/console/src/main/java/org/apache/syncope/client/console/panels/HistoryConfList.java#L52-L58).

ilgrosso: I see, thanks.
Anyway, as I don't think that an infinite drop-down is applicable anyway, can't we just always used the paged method display only the last N items, where N could be sensibly set to, say, 15/20?

About the existing method reference, think we are going to refactor / cleanup that too.

mmoayyed: Sounds good, sure; I'll make some adjustments.

client/console/src/main/java/org/apache/syncope/client/console/audit/AnyTOAuditEntryBean.java
```diff
@@ -0,0 +1,215 @@
...
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+import java.io.Serializable;
+import java.util.Date;
+import java.util.List;
+
+public class AnyTOAuditEntryBean implements Serializable {
```
ilgrosso: What's the need of `AnyTOAuditEntryBean`? Why `AuditEntryTO` is not enough?

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,316 @@
...
+     * @param anyTypeKind the any type kind
+     * @param anyTO       the any to
+     * @return the response
+     */
+    private ProvisioningResult<? extends AnyTO> restore(final AnyTOAuditEntryBean entryBean,
+                                                        final AnyTypeKind anyTypeKind, final AnyTO anyTO) {
+        try {
+            if (anyTypeKind == AnyTypeKind.USER) {
+                UserTO userTO = (UserTO) MAPPER.readValue(entryBean.getBefore().toString(), anyTypeKind.getTOClass());
+                UserTO current = new UserRestClient().read(anyTO.getKey());
```
ilgrosso: If comparison was shown against `anyTO`, why patch is generated against a freshly read object?

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,316 @@
...
+                userTO.setPassword(((UserTO) anyTO).getPassword());
```
ilgrosso: This statement should be removed - see below.

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,316 @@
...
+                userTO.setSecurityAnswer(((UserTO) anyTO).getSecurityAnswer());
```
ilgrosso: This statement should be removed - see below.

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,316 @@
...
+                UserPatch userPatch = AnyOperations.diff(userTO, current, false);
```
ilgrosso: After obtaining `userPatch`, add the following statements:

```java
userPatch.setPassword(null);
userPatch.setSecurityAnswer(null);
```

to ensure that no security-sensible changes are actually performed during restore.

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
```diff
@@ -0,0 +1,316 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class AuditHistoryDirectoryPanel extends
+    DirectoryPanel<AnyTOAuditEntryBean, AnyTOAuditEntryBean,
+        DirectoryDataProvider<AnyTOAuditEntryBean>, AuditHistoryRestClient>
```
ilgrosso: Replace `DirectoryDataProvider<AnyTOAuditEntryBean>` with `AuditHistoryProvider` here (need to make the subclass `protected`) - to align with other directory panels.

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryDirectoryPanel.java
mmoayyed: In reviewing this, the problem I ran into was the mismatched eTag value for update calls. When I used the `anyTO` object's eTag value for the update, I receive an exception for mismatched eTag values. If I use a freshly read object's eTag value, then the call passes through. In looking through the codebase, I found an [example here](https://github.com/apache/syncope/blob/2_1_X/client/console/src/main/java/org/apache/syncope/client/console/panels/LinkedAccountModalPanel.java#L177-L181) that more or less does the same I think. What might be a better alternative?

ilgrosso: Thanks for pointing that out the existing example, which is wrong and needs to be fixed - [noted](https://issues.apache.org/jira/browse/SYNCOPE-957?focusedCommentId=16986860&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16986860).

Running into mismatched eTag value errors is something that can simply happen - possibly because someone else has just updated the same user.
From the other side, what if one thinks to restore a certain version but the effective patch sent is calculated against the latest update that someone else has just sent?

You should simply pass the `getETagValue()` from original `anyTO` as [ordinarily performed](https://github.com/apache/syncope/blob/2_1_X/client/console/src/main/java/org/apache/syncope/client/console/wizards/any/UserWizardBuilder.java#L84) during update from Admin Console.

ilgrosso: One additional thing you could do is to take the updated ETag from `update()`'s response and pass it to existing `anyTO`'s `setEtagValue()`.

mmoayyed:
I have done some additional cleanup, and am going to test the create/update scenarios some more tomorrow.

client/console/src/main/java/org/apache/syncope/client/console/wicket/markup/html/form/AjaxDropDownChoicePanel.java
```diff
@@ -66,7 +66,7 @@ protected void onUpdate(final AjaxRequestTarget target) {
                 }
             });
         }
-
+
```
ilgrosso: Please remove this formatting-only change, thanks.

mmoayyed:
I have added a few more adjustments to clean up code and have also work through the NPE issues, etc. Tested the batch with created-only audit entries and those with history. Comparisons/updates also should now sanitize sensitive data for user objects, and use the correct object for diffs.

core/persistence-jpa-json/src/test/resources/domains/MasterContent.xml
```diff
@@ -2409,59 +2409,4 @@ $$ }&#10;
   <SecurityQuestion id="887028ea-66fc-41e7-b397-620d7ea6dfbb" content="What's your mother's maiden name?"/>
 
   <SyncopeLogger logName="syncope.audit.[LOGIC]:[SyncopeLogic]:[]:[isSelfRegAllowed]:[SUCCESS]" logLevel="DEBUG" logType="AUDIT"/>
-
```
ilgrosso: You removed all audit entries, not just read / list / search as I was suggesting... any reason why?

ilgrosso: Is `AuditITCase` working now, for example?

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryModal.java
```diff
@@ -0,0 +1,58 @@
...
+
+        super(BaseModal.CONTENT_ID);
+
+        final MultilevelPanel mlp = new MultilevelPanel("history");
+        mlp.setOutputMarkupId(true);
+        this.directoryPanel = getDirectoryPanel(mlp, baseModal, pageReference, entity);
+        add(mlp.setFirstLevel(this.directoryPanel));
+    }
+
+    protected DirectoryPanel<AuditEntryTO, AuditEntryTO,
```
ilgrosso: Please replace return type with `AuditHistoryDirectoryPanel`

client/console/src/main/java/org/apache/syncope/client/console/audit/AuditHistoryModal.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.syncope.client.console.wicket.markup.html.bootstrap.dialog.BaseModal;
+import org.apache.syncope.common.lib.to.AnyTO;
+import org.apache.syncope.common.lib.to.AuditEntryTO;
+import org.apache.wicket.PageReference;
+import org.apache.wicket.markup.html.panel.Panel;
+
+public class AuditHistoryModal<T extends AnyTO> extends Panel implements ModalPanel {
+    private static final long serialVersionUID = 1066124171682570080L;
+
+    protected final DirectoryPanel<AuditEntryTO, AuditEntryTO, ?, ?> directoryPanel;
```
ilgrosso: Please change type to `AuditHistoryDirectoryPanel`

core/persistence-jpa-json/src/test/resources/domains/MasterContent.xml
mmoayyed: ah, sorry my bad. Read your comment a little too quickly! I will clean this up to only remove those entries.

mmoayyed: Also I think I'd need to also include the same batch of audit loggers for AnyObjects and add test cases for those too.

ilgrosso: Ok; don't forget to align this file with `core/persistence-jpa-json/src/test/resources/domains/MasterContent.xml`

fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuditITCase.java
```diff
@@ -18,26 +18,28 @@
...
+import org.apache.syncope.common.lib.to.PagedResult;
 import org.apache.syncope.common.lib.to.UserTO;
 import org.apache.syncope.common.lib.types.AuditElements;
+import org.apache.syncope.common.rest.api.beans.AnyQuery;
 import org.apache.syncope.common.rest.api.beans.AuditQuery;
 import org.apache.syncope.core.persistence.api.entity.anyobject.AnyObject;
 import org.apache.syncope.fit.AbstractITCase;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.*;
```
ilgrosso: Please avoid star imports; possibly, restore the previous imports layout.

ilgrosso: Thanks!

mmoayyed:
This is G2G.

ilgrosso:
@mmoayyed LGTM please go ahead, squash and merge

PR #146, opened 2019-12-11
Replace JSTL with Apache Standard Taglibs

coheigea:
The JSTL dependency is not needed and comes with a CVE as it ships with an old version of Apache TagLibs.

### DIFF
--- a/ext/oidcclient/agent/pom.xml
+++ b/ext/oidcclient/agent/pom.xml
@@ -47,8 +47,8 @@ under the License.
       <artifactId>javax.servlet.jsp-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>javax.servlet</groupId>
-      <artifactId>jstl</artifactId>
+      <groupId>org.apache.taglibs</groupId>
+      <artifactId>taglibs-standard-impl</artifactId>
       <scope>provided</scope>
     </dependency>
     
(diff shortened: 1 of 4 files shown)


ilgrosso:
`jstl` is currently used by

* https://github.com/apache/syncope/blob/master/ext/saml2sp/agent/src/main/resources/META-INF/resources/saml2sp/loginSuccess.jsp
* https://github.com/apache/syncope/blob/master/ext/oidcclient/agent/src/main/resources/META-INF/resources/oidcclient/loginSuccess.jsp

hence if you want to remove it, you'll have to include in this PR the changes for such two JSPs.

Moreover, you should also include removal of https://github.com/apache/syncope/blob/master/standalone/LICENSE#L931

ilgrosso:
Moreover, scope is `provided`, hence nothing is actually shipped with release artifacts, no?

coheigea:
Maybe a better fix then would be to replace it with taglibs-standard-impl (leaving it as provided)?

coheigea:
@ilgrosso OK I modified the PR, please check again

ilgrosso:
LGTM.
Please go ahead, squash and merge.

coheigea:
@ilgrosso Thanks. I'll leave it up to you if you want to cherry-pick it or not

ilgrosso:
Thanks @coheigea ; cherry-picked with 469c3b30cd1cabeec6869a022a31c2d51c158c8a on `2_1_X` and ffb62885224d9eb680171f9454113bb92851dc0b on `2_0_X`

Please notice that on both branches JSTL is effectively included in the artifacts.

PR #161, opened 2020-01-20
Add validator for username field in admin console

mmoayyed:


ilgrosso:
I don't think it is a good idea to duplicate the validation check - which is performed by the Core for all entities, including User.
Furthermore, please consider that username values can be invalid not only because of illegal characters, but also because of one or more Account Policies defined, which can't be evaluated by Admin Console anyway.

ilgrosso:
Anyway, I have opened SYNCOPE-1534 for similar purpose.

PR #162, opened 2020-01-29
Check for presence of value before using an Optional

mmoayyed:
This pull request:

- Removes a number of unused imports from Syncope
- Primarily makes sure, in certain areas, that Optional objects are checked to actually contain a value before a value can be retrieved. I specifically ran into an example of this, as a result of bad (manual) mapping configuration in the MasterContent.xml file, where no mapping was tagged with a `conn-object-key=1`. I decided to add a bit more safety to protect and defect against the dark arts...of dealing with optionals! :)

### DIFF
--- a/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
+++ b/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
@@ -51,6 +51,7 @@ import org.apache.syncope.core.persistence.api.dao.AccessTokenDAO;
 import org.apache.syncope.core.persistence.api.dao.ConfDAO;
 import org.apache.syncope.core.persistence.api.dao.SecurityQuestionDAO;
 import org.apache.syncope.core.persistence.api.entity.group.Group;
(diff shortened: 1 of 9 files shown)


core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -414,8 +415,10 @@ private boolean isPasswordMapped(final ExternalResource resource) {
                     ResourceOperation.UPDATE,
                     anyUtils.getAllResources(toBeUpdated).stream().
                             filter(resource -> resource.getProvision(toBeUpdated.getType()).isPresent()).
-                            filter(resource -> mappingManager.hasMustChangePassword(
-                            resource.getProvision(toBeUpdated.getType()).get())).
+                            filter(resource -> {
```
ilgrosso: Why isn't the first `filter(resource -> resource.getProvision(toBeUpdated.getType()).isPresent())` enough?

mmoayyed: Good point. I missed the filter actually and will revert this.

PR #174, opened 2020-04-14
SYNCOPE-1552: Allow WA audit events to be stored in core

mmoayyed:


wa/starter/pom.xml
```diff
@@ -43,13 +43,16 @@ under the License.
...
-
     <dependency>
       <groupId>org.apache.syncope.common.keymaster</groupId>
       <artifactId>syncope-common-keymaster-client-api</artifactId>
       <version>${project.version}</version>
     </dependency>
-
+    <dependency>
+      <groupId>org.apache.syncope.core</groupId>
+      <artifactId>syncope-core-provisioning-api</artifactId>
```
ilgrosso, line 53: `syncope-core-provisioning-api` are meant only for usage by `Core`: the relevant class(es) should be extracted and placed into `common/idrepo/lib`.

PR #176, opened 2020-04-21
SYNCOPE-1553: Translate AuthModuleConfs to WA

mmoayyed:
TODO:

- Translate profile item mappings for attributes, etc.
- Fix configuration data and tests for AuthModuleITs to not bootstrap the context with invalid state/configuration.

### DIFF
--- a/core/am/logic/src/main/java/org/apache/syncope/core/logic/AuthModuleLogic.java
+++ b/core/am/logic/src/main/java/org/apache/syncope/core/logic/AuthModuleLogic.java
@@ -59,8 +59,7 @@ public class AuthModuleLogic extends AbstractTransactionalLogic<AuthModuleTO> {
         return binder.getAuthModuleTO(authModuleDAO.save(binder.update(authModule, authModuleTO)));
     }
 
-    @PreAuthorize("isAnonymous() or hasRole('" + AMEntitlement.AUTH_MODULE_LIST
-        + "') or hasRole('" + IdRepoEntitlement.ANONYMOUS + "')")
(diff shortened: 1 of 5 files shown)


core/am/logic/src/main/java/org/apache/syncope/core/logic/AuthModuleLogic.java
```diff
@@ -60,7 +59,8 @@ public AuthModuleTO update(final AuthModuleTO authModuleTO) {
         return binder.getAuthModuleTO(authModuleDAO.save(binder.update(authModule, authModuleTO)));
     }
 
-    @PreAuthorize("hasRole('" + AMEntitlement.AUTH_MODULE_LIST + "')")
+    @PreAuthorize("isAnonymous() or hasRole('" + AMEntitlement.AUTH_MODULE_LIST
```
ilgrosso, line 62: Hum I don't see any reason to apply such a change: what's the rationale, please?

mmoayyed: The PropertySourceLocator fails to call `list()` on AuthModuleService via the anonymous-user, and produces 403 error codes. Cross-checked with 1-2 other Logic components and they seem to follow a similar approach. Is there a better alternative?

ilgrosso: I see, then I'd remove `isAnonymous()` from the list, thus leaving the condition "`IdRepoEntitlement.ANONYMOUS` or `AMEntitlement.AUTH_MODULE_LIST`"

mmoayyed:
PS Just a note for public's and archive's sake that the saml stuff likely will be revised shortly to handle mappings better with richer pac4j support.

PR #178, opened 2020-04-28
[SYNCOPE-1555] Allow WA as SAMLs IdP to fetch metadata over REST

mat-ale:


common/am/lib/src/main/java/org/apache/syncope/common/lib/to/SAML2IdPMetadataTO.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    @Override
+    public String getKey() {
+        return key;
+    }
+
+    @Override
+    public void setKey(final String key) {
```
ilgrosso, line 92: This methods needs to be further annotated with `@PathParam("key")` because it is used as an argument of `SAML2IdPMetadataConfService#update`

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/auth/JPASAML2IdPMetadataDAO.java
```diff
@@ -0,0 +1,61 @@
...
+        TypedQuery<SAML2IdPMetadata> query = entityManager().createQuery(
+                "SELECT e FROM " + JPASAML2IdPMetadata.class.getSimpleName() + " e WHERE e.appliesTo=:appliesTo",
+                SAML2IdPMetadata.class);
+        query.setParameter("appliesTo", appliesTo);
+
+        SAML2IdPMetadata result = null;
+        try {
+            result = query.getSingleResult();
+        } catch (final NoResultException e) {
+            LOG.info("No SAML2 IdP Metadata found with appliesTo = {}", appliesTo);
```
ilgrosso, line 49: This should be `LOG.debug()`

fit/core-reference/pom.xml
```diff
@@ -222,6 +222,30 @@ under the License.
       <artifactId>bcpkix-jdk15on</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
ilgrosso, line 225: Please remove all these added dependencies

wa/starter/src/main/java/org/apache/syncope/wa/SyncopeCoreNotReadyException.java
```diff
@@ -0,0 +1,25 @@
+/*
```
ilgrosso, line 1: Is this exception really required? How are existing components in WA managing the fact that WARestClient is  not ready yet?

fit/core-reference/pom.xml
mat-ale: wow I did not add those depedencies, removing!

wa/starter/src/main/java/org/apache/syncope/wa/SyncopeCoreNotReadyException.java
mat-ale: I think @mmoayyed will be able to help us about this question!

mat-ale: ATM in the other places where we check whether the client is ready or not we just have logging statements, no exceptions!

mmoayyed: Strictly speaking, this is not required here because we'd be making an assumption that core is in fact ready by the time this method is called (and is only called once the app is ready). If by then core is not ready, we should likely consider that a catastrophic event and end the app, specially because this (metadata, etc) is not the sort of thing one can ignore unlike other places. So perhaps we can just return `false` here, but...

..more importantly, I see the catch clause that traps `Exception`, and by extension this one, just logs a warning. I don't think that's right; at the point in the lifetime of the app when this method is called, core must be ready. Its unavailability should halt the app and present an error (more than a warning).

mat-ale: @mmoayyed What do you mean with "So perhaps we can just return `false` here"?
According to your clarification I guess that one option could be to pull out the `getSyncopeClient()` from the `try/catch` and make it throw the blocking exception in order to block the execution.
But about the `SyncopeCoreNotReadyException`, do you guys want to keep it or just throw a `RuntimeException` with `Syncope core is not yet ready to access requests` as message?

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/auth/JPASAML2IdPMetadataDAO.java
```diff
@@ -0,0 +1,61 @@
...
+        try {
+            result = query.getSingleResult();
+        } catch (final NoResultException e) {
+            LOG.debug("No SAML2 IdP Metadata found with appliesTo = {}", appliesTo);
+        }
+
+        return result;
+    }
+
+    @Transactional(readOnly = true)
```
mmoayyed, line 55: Should this be readonly? Seems like this might be more appropriate for other findXYZ methods?

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/SAML2IdPMetadataBinderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+            final SAML2IdPMetadataTO saml2IdPMetadataTO) {
+
+        SAML2IdPMetadata result = saml2IdPMetadata;
+
+        if (result == null) {
+            result = entityFactory.newEntity(SAML2IdPMetadata.class);
+        }
+
+        SAML2IdPMetadata saml2IdPMetadataResult = SAML2IdPMetadata.class.cast(result);
+        SAML2IdPMetadataTO saml2IdPMetadataTOElem = SAML2IdPMetadataTO.class.cast(saml2IdPMetadataTO);
```
mmoayyed, line 45: Is this cast necessary?

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/SAML2IdPMetadataBinderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+        saml2IdPMetadataResult.setSigningCertificate(saml2IdPMetadataTOElem.getSigningCertificate());
+        saml2IdPMetadataResult.setSigningKey(saml2IdPMetadataTOElem.getSigningKey());
+        saml2IdPMetadataResult.setAppliesTo(saml2IdPMetadataTOElem.getAppliesTo());
+
+        return result;
+    }
+
+    @Override
+    public SAML2IdPMetadata create(final SAML2IdPMetadataTO saml2IdPMetadataTO) {
+        return getSAML2IdPMetadata(null, saml2IdPMetadataTO);
```
mmoayyed, line 59: Instead of null, could we pass `entityFactory.newEntity(SAML2IdPMetadata.class);`?

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/auth/JPASAML2IdPMetadataDAO.java
mat-ale: yes that's right thanks, I am gonna move it to `#findByOwner` and `#find` methods!

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/SAML2IdPMetadataBinderImpl.java
mat-ale: no right, not necessary! Gonna remove it

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/SAML2IdPMetadataBinderImpl.java
mat-ale: yes correct, we normally do like you said, elsewhere! Gonna fix it

ilgrosso: @mat-ale if you mean renaming that method to `update()` as other data binders do, yes please go on.

PR #179, opened 2020-05-01
Bumping year to 2020

coheigea:


ilgrosso:
@coheigea thanks for reporting this!

There are, however, several `NOTICE` files in the source tree that should be amended as well.

Moreover, we'll need to do the same with branches `2_1_X` and `2_0_X`

coheigea:
@ilgrosso I updated the PR to pick up the other NOTICE files

ilgrosso:
@coheigea by looking at other years' similar commits I found the following file should be updated as well: 'src/site/site.xml`

coheigea:
@ilgrosso Done! Could you backport to the older branches?

ilgrosso:
@coheigea sure

PR #182, opened 2020-05-04
NOJIRA - Switch Travis CI to AMD64 Bionic Distro

mmoayyed:
The Travis CI trusty image is deprecated. This change switches the distro to Bionic which builds against Ubuntu 18.

PS. I also experimented with arm-based builds which are super fast to startup (less than 10 seconds) but they are not as packed as others (No maven, Java installed, etc) and the cache seems somewhat suspicious with those. Bionic should do fine for now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2530,7 +2530,6 @@ under the License.
             <exclude>**/*.json</exclude>
             <exclude>**/banner.txt</exclude>
             <exclude>**/target/**</exclude>
-            <exclude>**/install-jdk.sh</exclude>
           </excludes>
         </configuration>
         <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -2530,6 +2530,7 @@ under the License.
             <exclude>**/*.json</exclude>
             <exclude>**/banner.txt</exclude>
             <exclude>**/target/**</exclude>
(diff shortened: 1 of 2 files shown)


ilgrosso:
@mmoayyed it seems that installing other JDKs is failing, I tried to restart the related steps several times, no luck. Any idea?

pom.xml
```diff
@@ -2530,6 +2530,7 @@ under the License.
             <exclude>**/*.json</exclude>
             <exclude>**/banner.txt</exclude>
             <exclude>**/target/**</exclude>
+            <exclude>**/install-jdk.sh</exclude>
```
ilgrosso, line 2533: I think we should avoid this.
Isn't enough to fetch the `install-jdk` script into a different location, not placed in the checkout directory, to avoid RAT failing?

mmoayyed: Ah, that's much better. I'll see if I can fetch it into the temp directory.

ilgrosso: maybe `~/bin` as it used to be?

ilgrosso:
@mmoayyed I'd say it works, please go ahead, squash and merge, thanks.

PR #183, opened 2020-05-06
NOJIRA: skipTests maven profile should also apply to ITs

mmoayyed:


ilgrosso:
@mmoayyed any reason why you need this?

mmoayyed:
I would like `clean install -PskipTests`s to complete as quickly as possible, skipping as many tests as possible. It also seems strange that I can specify skipping tests, but in fact, only certain type of tests are skipped. Skipping tests should mean skipping tests, no surprises :)

ilgrosso:
I see, but when using `-PskipTests` IT Cases are already not executed, even without the change proposed in this PR.

mmoayyed:
hmmm, that's odd. I wonder if this may be caused by a conflict between the build and my local maven settings. I do see ITs run when I run with that profile, using `-PskipTests,all`. I will remove/reset local changes to try again. Will report back shortly.

ilgrosso:
This is so true that we use such profile for initial Travis CI step: https://github.com/apache/syncope/blob/master/.travis.yml#L51

mmoayyed:
Tested again; Seems like I might have had some local profile changes messing up the build. Change is unnecessary as was pointed out. Closing. Thanks!

PR #184, opened 2020-05-06
[SYNCOPE-1557] Allowing for flexible Auth, Access, AttrRelease and ClientApp translation

ilgrosso:


wa/starter/src/main/java/org/apache/syncope/wa/starter/mapping/AllowedAttrReleaseMapper.java
```diff
@@ -0,0 +1,55 @@
...
+        registeredServiceAttributeReleasePolicyClass = ReturnAllowedAttributeReleasePolicy.class)
+@Component
+public class AllowedAttrReleaseMapper implements AttrReleaseMapper {
+
+    @Override
+    public RegisteredServiceAttributeReleasePolicy build(final AttrReleasePolicyConf conf) {
+        RegisteredServiceAttributeReleasePolicy attributeReleasePolicy;
+        if (!((AllowedAttrReleasePolicyConf) conf).getAllowedAttrs().isEmpty()) {
+            attributeReleasePolicy = new ReturnAllowedAttributeReleasePolicy();
+            ((AllowedAttrReleasePolicyConf) conf).getAllowedAttrs();
```
mmoayyed, line 39: This seems like a dangling line?

wa/starter/src/main/java/org/apache/syncope/wa/starter/mapping/RegisteredServiceMapper.java
```diff
@@ -0,0 +1,161 @@
...
+        });
+
+        ctx.getBeansOfType(ClientAppMapper.class).forEach((name, bean) -> {
+            ClientAppMapFor clientAppMapFor = ctx.findAnnotationOnBean(name, ClientAppMapFor.class);
+            if (clientAppMapFor != null) {
+                clientAppTOMappers.put(clientAppMapFor.clientAppClass().getName(), bean);
+                registeredServiceMappers.put(clientAppMapFor.registeredServiceClass().getName(), bean);
+            }
+        });
+    }
```
mmoayyed, line 97: I think it would be better if all of this was moved to the configuration class that creates the bean instance, removing the need to implement an interface and inject an application context. Injecting the application context is a discouraged practice, and should be avoided when possible. It also makes it more difficult to write unit tests for this.

PR #188, opened 2020-05-15
[SYNCOPE-1563] Fixed password mangement in user approval form

andrea-patricelli:
Do not initialize user approval form with the password taken from workflow


ext/flowable/client-console/src/main/java/org/apache/syncope/client/console/panels/UserRequestFormDirectoryPanel.java
```diff
@@ -226,8 +226,10 @@ public void onClick(final AjaxRequestTarget target, final UserRequestForm ignore
                 UserRequestForm formTO = model.getObject();
                 UserTO newUserTO;
                 UserTO previousUserTO;
-                if (formTO.getUserPatch() == null) {
+                if (formTO.getUserPatch() == null && formTO.getUserTO() != null) {
```
ilgrosso, line 229: Maybe you can move this check right before the line below, e.g.

```java
if (newUserTO != null) {
  // SYNCOPE-1563 do not use the password into UserTO
  newUserTO.setPassword(null);
}
```

PR #195, opened 2020-06-08
Fixing a grammatical issue with the mail template

coheigea:
"was request for" -> "was requested for"


ilgrosso:
Thanks @coheigea !

PR #196, opened 2020-06-10
SYNCOPE-1556: Manage OIDC JWKS

mmoayyed:
- Allows Syncope to manage the JWKS store via REST endpoints
- Uses the `nimbus-jose-jwt` library to generate keys with algorithm RSA and length 2048
- Provides two sets of APIs; one to be invoked by the admin console for delete/update tasks and the other may be used by WA. (Permissions are kept separate)
- Adds tests and integration tests.

### DIFF
--- a/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/auth/JPAOIDCJWKSDAO.java
+++ b/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/auth/JPAOIDCJWKSDAO.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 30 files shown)


core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/auth/JPAOIDCJWKSDAO.java
```diff
@@ -0,0 +1,60 @@
...
+@Repository
+public class JPAOIDCJWKSDAO extends AbstractDAO<OIDCJWKS> implements OIDCJWKSDAO {
+
+    @Transactional(readOnly = true)
+    @Override
+    public Optional<OIDCJWKS> get() {
+        try {
+            TypedQuery<OIDCJWKS> query = entityManager().
+                createQuery("SELECT e FROM " + JPAOIDCJWKS.class.getSimpleName() + " e", OIDCJWKS.class);
+            return Optional.of(query.getSingleResult());
```
ilgrosso, line 42: The equivalent methods from other DAOs are not returning `Optional` but something that can be `null`.
I recognize the advantages of such approach, but I think we should also keep consistency.

We could consider a future PR involving all DAOs.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/OIDCJWKSDataBinderImpl.java
```diff
@@ -0,0 +1,68 @@
...
+            build();
+    }
+
+    @Override
+    public OIDCJWKS create() {
+        try {
+            OIDCJWKS jwks = entityFactory.newEntity(OIDCJWKS.class);
+            RSAKey jwk = new RSAKeyGenerator(2048)
+                .keyUse(KeyUse.SIGNATURE)
+                .keyID(UUID.randomUUID().toString())
```
ilgrosso, line 54: Do not use `UUID.randomUUID().toString()` but `SecureRandomUtils.generateRandomUUID()`

PR #202, opened 2020-07-10
NOJIRA: Allow push ops to WA to reload services

mmoayyed:


common/am/rest-api/src/main/java/org/apache/syncope/common/rest/api/service/ClientAppService.java
```diff
@@ -131,4 +131,11 @@
     @Path("{type}/{key}")
     @Produces({ MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML })
     void delete(@NotNull @PathParam("type") ClientAppType type, @NotNull @PathParam("key") String key);
+
+    @ApiResponses(
+        @ApiResponse(responseCode = "204", description = "Operation was successful"))
+    @POST
+    @Path("push")
+    @Produces({ MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML })
+    void push();
```
ilgrosso, line 140: Would you mind renaming this method as `pushToWA`, similar to its equivalent in `SRARouteService#pushToSRA`?

PR #203, opened 2020-07-13
SYNCOPE-1577: Support CAS protocol for client applications

mmoayyed:
Cf. https://issues.apache.org/jira/browse/SYNCOPE-1577


fit/core-reference/src/test/java/org/apache/syncope/fit/AbstractITCase.java
```diff
@@ -820,6 +821,31 @@ protected void removeLdapRemoteObject(
         return object;
     }
 
+    protected CASSPTO buildCASSP() {
```
ilgrosso, line 824: Maybe this method can be moved to `ClientAppITCase` as no one else is using it ATM

PR #204, opened 2020-07-15
SYNCOPE-1545: Enable Swagger support for WA

mmoayyed:
This pull request adjusts WA to turn on Swagger support via the `cas-server-support-swagger` dependency. In doing so,

1. The [CAS module is upgraded](https://apereo.github.io/cas/development/integration/Swagger-Integration.html) to base its swagger support on top of Swagger v2.1.x
2. The WA configuration does a simple bean override to present Syncope information to the Docklet, such as version, title, etc.
3. Default configuration is adjusted to consider actuators, controllers etc as part of Swagger UI.

image

### DIFF
--- a/wa/starter/src/main/java/org/apache/syncope/wa/starter/config/SyncopeWAConfiguration.java
+++ b/wa/starter/src/main/java/org/apache/syncope/wa/starter/config/SyncopeWAConfiguration.java
@@ -74,8 +74,6 @@ import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
-import javax.annotation.Resource;
-
 import java.time.LocalDate;
 import java.time.ZoneId;
 import java.util.Collection;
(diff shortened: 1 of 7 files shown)


wa/starter/src/main/java/org/apache/syncope/wa/starter/config/SyncopeWAConfiguration.java
```diff
@@ -221,4 +237,14 @@ public KeymasterStart keymasterStart() {
...
         return new KeymasterStop(NetworkService.Type.WA);
     }
+
+    @Bean
+    public String version() {
+        return applicationContext.getEnvironment().getProperty("version");
+    }
+
+    @Bean
+    public String buildNumber() {
```
ilgrosso, line 247: This should not be needed.
Please also consider that it would require the `buildnumber-maven-plugin` to run.

wa/starter/src/main/resources/application.properties
```diff
@@ -45,3 +45,6 @@ spring.main.allow-bean-definition-overriding=true
 spring.main.lazy-initialization=false
 
 service.discovery.address=http://localhost:8080/syncope-wa/
+
+version=${syncope.version}
+buildNumber=${buildNumber}
```
ilgrosso, line 50: see above

ilgrosso:
@mmoayyed looks very nice, thanks!
Besides the small things noted above, would you mind, when merging, to add `SYNCOPE-1545` in the commit message? Thanks.

mmoayyed:
Certainly! Will do.

PR #205, opened 2020-07-15
[SYNCOPE-1578] Adding auth support to SRA, with OIDC / OAUTH2 implementations, revie…

ilgrosso:
…wing several details on WA and Core to make all functional


ilgrosso:
This looks somewhat stuck, closing and creating new one.

PR #207, opened 2020-07-16
SYNCOPE-1580: Manage WA Configuration Properties/Schema

mmoayyed:
See: https://issues.apache.org/jira/browse/SYNCOPE-1580

- Provides a `WAConfigService`, along with relevant logic, dao and binder components to CRUD configuration entries. The caller of `WAConfigService` is required to push changes to WA once entries are stored, allowing WA to refresh itself, as it does today with authentication modules and the bootstrap process. `WAConfigService` provides for a flexible and generic way to manage *all* properties relevant to a Spring Boot application, whether they belong to CAS or some other specific Spring Boot setting.

- Presents a `WAConfigTO`, backed by the `WAConfigEntry` at the JPA layer to assist with `WAConfigService` implementation. The value of a configuration entry is supported by generics to allow creation/fetching of generic types for values (lists, numbers, etc)

- WA can only fetch/read/list configuration entries (i.e. anonymous is allowed). All other non-readonly operations are adjusted with the right permissions.

- Bootstrapping WA properties is now handled via special filters in type-safe maintainable ways (i.e no Strings to match property keys/names)that attempt to only fetch and collect properties supported by WA (and not everything supported by CAS properties). This means the collection of properties are either retrieved from authentication module configurations (LDAP, Syncope, etc) or from `WAConfigService` for things that may not be related to an authentication module (i.e. OIDC issuer field).


core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/auth/WAConfigEntry.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.apache.syncope.core.persistence.api.entity.auth;
+
+import org.apache.syncope.core.persistence.api.entity.ProvidedKeyEntity;
+
+import java.io.Serializable;
+
+public interface WAConfigEntry<T extends Serializable> extends ProvidedKeyEntity {
+
+    String getName();
```
ilgrosso, line 27: What is the need for `name`? Since `key` is to be provided (not generated, correctly), `name` seems to be just a duplicate, no?

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/auth/JPAWAConfigEntry.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Lob;
+import javax.persistence.Table;
+
+import java.io.Serializable;
+
+@Entity
+@Table(name = JPAWAConfigEntry.TABLE)
+public class JPAWAConfigEntry extends AbstractGeneratedKeyEntity implements WAConfigEntry {
```
ilgrosso, line 33: Since `WAConfigEntry` extends `ProvidedKeyEntity`, this class must extend `AbstractProvidedKeyEntity`.

As said above, `name` should be removed.

core/persistence-jpa/src/test/java/org/apache/syncope/core/persistence/jpa/inner/WAConfigTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    public void save() {
+        create("system.example.key[0]", "value1,value2,value3");
+        assertFalse(configDAO.findAll().isEmpty());
+    }
+
+    @Test
+    public void update() {
+        WAConfigEntry entry = create("system.syncope.key[0]", "value1,value2,value3");
```
ilgrosso, line 54: Please add tests for more value types.

fit/core-reference/src/test/java/org/apache/syncope/fit/core/WAConfigITCase.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class WAConfigITCase extends AbstractITCase {
+    @Test
+    public void verify() {
+        WAConfigTO configTO = new WAConfigTO.Builder()
```
ilgrosso, line 39: Please add tests for more value types.

common/am/rest-api/src/main/java/org/apache/syncope/common/rest/api/service/wa/WAConfigService.java
```diff
@@ -0,0 +1,114 @@
...
+    @Path("{key}")
+    @Consumes({MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML})
+    @Produces({MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML})
+    void update(@NotNull WAConfigTO configTO);
+
+    @DELETE
+    @Consumes({MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML})
+    @Produces({MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML})
+    @Path("{key}")
+    Response delete(@NotNull @PathParam("key") String key);
```
ilgrosso, line 106: Please change return type from `Response` to `void`

ilgrosso:
@mmoayyed it seems there are type troubles with XML: https://travis-ci.org/github/apache/syncope/jobs/710281684#L703

ilgrosso:
@mmoayyed still errors with XML, it seems: https://travis-ci.org/github/apache/syncope/jobs/710862476#L931

mmoayyed:
Sure. What would be a good substitute for the logic bounds?

```
class WAConfigLogic extends AbstractTransactionalLogic<WAConfigTO>
```

Can I use the following?

```
class WAConfigLogic extends AbstractTransactionalLogic<EntityTO>
```

ilgrosso:
@mmoayyed I see two options: one, the simpler, is what you suggest, e.g. extending `AbstractTransactionalLogic<EntityTO>`; the other is to define
```java 
public class WAConfigTO extends Attr implements EntityTO
```

PR #213, opened 2020-09-08
SYNCOPE-1592: Propagation task executions and no-SQL persistence

andrea-patricelli:
This is the fog-catcher implementation for "Propagation task executions and no-SQL persistence", that includes:

- Refactoring of persistence layer to have a table for each task type (push, pull, propagation, notification, sched) with related SQL script to upgrade from previous versions.
- Persistence of PropagationTasks on Elasticsearch as per any objects.


client/console/src/main/resources/org/apache/syncope/client/console/panels/GroupDirectoryPanel_fr_CA.properties
```diff
@@ -16,3 +16,4 @@
 # under the License.
 any.edit=Modifier ${anyTO.type} ${anyTO.name}
 group.members=${right} membres de ${left.name}
+auditHistory.title=${anyTO.type} ${anyTO.fullGroupName != "-" ? anyTO.fullGroupName : anyTO.name} histoire
```
ilgrosso, line 19: Please do a straight commit for this fix, both for `2_1_X` and `master`, no need to include it in this PR.

core/logic/src/main/java/org/apache/syncope/core/logic/SyncopeLogic.java
```diff
@@ -448,7 +448,7 @@ public NumbersInfo numbers() {
         numbersInfo.getConfCompleteness().put(
                 NumbersInfo.ConfItem.NOTIFICATION.name(), !notificationDAO.findAll().isEmpty());
         numbersInfo.getConfCompleteness().put(
-                NumbersInfo.ConfItem.PULL_TASK.name(), !taskDAO.findAll(TaskType.PULL).isEmpty());
```
ilgrosso, line 451: Did you remove the `TaskType` enum completely, right?

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPATaskDAO.java
```diff
@@ -167,11 +141,11 @@ private String getEntityTableName(final TaskType type) {
...
     }
 
-    private <T extends Task> StringBuilder buildFindAllQueryJPA(final TaskType type) {
+    private <T extends Task> StringBuilder buildFindAllQueryJPA(final Class<T> reference) {
         StringBuilder builder = new StringBuilder("SELECT t FROM ").
-                append(getEntityReference(type).getSimpleName()).
+                append(getEntityReference(reference).getSimpleName()).
                 append(" t WHERE ");
-        if (type == TaskType.SCHEDULED) {
+        if (SchedTask.class.isAssignableFrom(reference)) {
```
ilgrosso, line 148: This is not correct, as `type == TaskType.SCHEDULE` used to match only `SchedTask` while now this is also matching `PushTask`, `PullTask` and so on.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPATaskDAO.java
```diff
@@ -216,84 +190,86 @@ private StringBuilder buildFindAllQuery(
...
+            queryString.
+                    append(", null AS startDate, null AS endDate, null AS status FROM ").
+                    append(tableName).
+                    append(" WHERE 1=1");
         }
 
-        queryParameters.add(getEntityTableName(type));
-        if (type == TaskType.SCHEDULED) {
+        queryParameters.add(getEntityTableName(reference));
+        if (SchedTask.class.isAssignableFrom(reference)) {
```
ilgrosso, line 238: Same here as above.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPATaskDAO.java
```diff
@@ -436,7 +413,13 @@ public int count(
     @Transactional(rollbackFor = { Throwable.class })
     @Override
     public <T extends Task> T save(final T task) {
-        return entityManager().merge(task);
+        T merged = entityManager().merge(task);
+
+        // propagate the event only for Propagation tasks
+        if (merged instanceof PropagationTask) {
+            publisher.publishEvent(new TaskCreatedUpdatedEvent<>(this, merged, AuthContextUtils.getDomain()));
```
ilgrosso, line 420: Maybe `TaskCreatedUpdatedEvent` shall be renamed as `PropagationTaskCreatedUpdatedEvent`?

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPATaskDAO.java
```diff
@@ -458,11 +441,12 @@ public void delete(final Task task) {
         }
 
         entityManager().remove(task);
+        publisher.publishEvent(new TaskDeletedEvent(this, task.getKey(), AuthContextUtils.getDomain()));
```
ilgrosso, line 444: Why not doing this only for propagation tasks, as done above, instead?

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/task/AbstractSchedTask.java
```diff
@@ -0,0 +1,121 @@
...
+import javax.persistence.Temporal;
+import javax.persistence.TemporalType;
+import javax.validation.constraints.NotNull;
+import org.apache.syncope.common.lib.types.ImplementationType;
+import org.apache.syncope.core.persistence.api.entity.Implementation;
+import org.apache.syncope.core.persistence.api.entity.task.SchedTask;
+import org.apache.syncope.core.persistence.jpa.entity.JPAImplementation;
+
+@MappedSuperclass
+public abstract class AbstractSchedTask extends AbstractTask implements SchedTask {
```
ilgrosso, line 33: Since `JPASchedTask` implements `SchedTask`, I think it would be cleaner to remove `implements SchedTask` from here.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/task/JPATaskExec.java
```diff
@@ -42,6 +43,7 @@
      * The referred task.
      */
     @ManyToOne(optional = false)
+    @JoinColumn(name = "task_id", nullable = false)
```
ilgrosso, line 46: Is this `@JoinColumn` annotation really needed? What if you remove it?

core/persistence-jpa/src/test/resources/domains/Master.properties
```diff
@@ -16,6 +16,7 @@
 # under the License.
 Master.driverClassName=org.h2.Driver
 Master.url=jdbc:h2:mem:syncopedb;DB_CLOSE_DELAY=-1
+#Master.url=jdbc:h2:file:~/syncopedb;DB_CLOSE_ON_EXIT=FALSE;AUTO_RECONNECT=TRUE
```
ilgrosso, line 19: Please revert this change.

fit/console-reference/pom.xml
```diff
@@ -276,6 +276,7 @@ under the License.
 
       <properties>
         <skipTests>true</skipTests>
+        <docker.autoCreateCustomNetworks>true</docker.autoCreateCustomNetworks>
```
ilgrosso, line 279: Please revert the changes to this file, they are not needed.

fit/core-reference/pom.xml
```diff
@@ -443,6 +443,74 @@ under the License.
   </build>
   
   <profiles>
+    <profile>
```
ilgrosso, line 446: Please revert the changes to this file, they are not needed.

ilgrosso (changes requested):
Besides the comments I have added, there are two additional things to cover:

1. run the tests for `core/persistence-jpa-json` will likely lead to failures, as the `MasterContent.xml` there was not updated
1. at least a test case should be added in `ElasticSearchITCase` to cover the new features

core/logic/src/main/java/org/apache/syncope/core/logic/SyncopeLogic.java
andrea-patricelli: Nope, going to remove it.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPATaskDAO.java
andrea-patricelli: Yes, I left the implementation in a way "open" also to other tasks. There are some cases in which would be useful also to have `NotificationTask` on ES.

ilgrosso: It does not make much sense to leave open about things that will never be invoked, as the only caller can realistically be `JPATaskDAO`, and a clear `if` is placed there around publisher invocation.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/task/AbstractSchedTask.java
andrea-patricelli: So each class that extends `AbstractSchedTask` has to define sched task (`JPAPullTask`, `JPAPushTask`, `JPASchedTask`) specific attributes and implement specific methods?

ilgrosso: No, my comment as above is just about removing `implements SchedTask ` from `AbstractSchedTask`, since, in the same inheritance tree, there is a subclass, e.g. `JPASchedTask` implementing the same interface - not very clean.

andrea-patricelli: Oh ok, now I got. Yes, I agree, I'll leave it in `AbstractSchedTask` and remove in `JPASchedTask.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPATaskDAO.java
andrea-patricelli: Ok, going to replace with `JPASchedTask`

core/logic/src/main/java/org/apache/syncope/core/logic/SyncopeLogic.java
andrea-patricelli: We cannot remove `TaskType` enum, it is necessary for search and to build `TaskQuery` objects.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/task/JPATaskExec.java
andrea-patricelli: Yes, this is needed. Without this the tests `org.apache.syncope.core.persistence.jpa.outer.TaskTest` are going to fail with errors like `Column "TASKEXEC.TASK_ID" not found;`.

andrea-patricelli:
> Besides the comments I have added, there are two additional things to cover:
> 
> 1. run the tests for `core/persistence-jpa-json` will likely lead to failures, as the `MasterContent.xml` there was not updated
This has been solved.
> 2. at least a test case should be added in `ElasticSearchITCase` to cover the new features
I thought that, as per user search, just current Task IT test were enough, since we simply provide a new implementation for task search.

ilgrosso:
About 1 year with no updates, let's close this PR.
It can be re-opened if needed.

PR #216, opened 2020-09-17
SYNCOPE-1589: WA WebAuthN Device Registration APIs

mmoayyed:
Allows WA to manage WebAuthN device registration requests. 

In this pull request, authentication profiles are adjusted to include a reference to a `WebAuthnRegisteredAccount` that itself contains a list of `WebAuthnDeviceCredential` entries, allowing the account management facility to track multiple devices for the same profile owner. The `WebAuthnDeviceCredential` component tracks a few key pieces related to the device, and also tracks the entire record as a JSON blob where this record is deserialized back into an object representation that can be used by WA.


core/am/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/wa/WebAuthnRegistrationServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+import java.net.URI;
+import java.util.List;
+
+@Service
+public class WebAuthnRegistrationServiceImpl extends AbstractServiceImpl implements WebAuthnRegistrationService {
+    @Autowired
+    private WebAuthnRegistrationServiceLogic logic;
+
+    @Override
+    public PagedResult<WebAuthnRegisteredAccount> list() {
```
ilgrosso, line 42: Since you're always setting page to `1`, just change this method's signature (in interface as well, naturally) to

```List<WebAuthnRegisteredAccount> list();```

File: common/am/lib/src/main/java/org/apache/syncope/common/lib/types/WebAuthnRegisteredAccount.java [hunk 2]
```diff
@@ -0,0 +1,124 @@
...
+package org.apache.syncope.common.lib.types;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.syncope.common.lib.BaseBean;
+
+import java.util.List;
+
+public class WebAuthnRegisteredAccount implements BaseBean {
```
ilgrosso, line 28: Any reason not to use the shorter `WebAuthnAccount` ubiquitously, instead?

common/am/rest-api/src/main/java/org/apache/syncope/common/rest/api/service/wa/WebAuthnRegistrationService.java
```diff
@@ -0,0 +1,102 @@
...
+    @DELETE
+    @Consumes({MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML})
+    @Produces({MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML})
+    @Path("${owner}/${credentialId}")
+    Response delete(@NotNull @PathParam("owner") String owner, @NotNull @PathParam("credentialId") String credentialId);
+
+    @DELETE
+    @Consumes({ MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML })
+    @Produces({ MediaType.APPLICATION_JSON, RESTHeaders.APPLICATION_YAML, MediaType.APPLICATION_XML })
+    Response deleteAll();
```
ilgrosso, line 84: Besides tests, is there any other place where such a method is used?

ilgrosso (changes requested):

common/am/rest-api/src/main/java/org/apache/syncope/common/rest/api/service/wa/WebAuthnRegistrationService.java
mmoayyed: At the moment this is only used for tests, and it does have the potential to act as public API if one were to try and delete all devices. I can definitely take it out if that is too dangerous, per se.

ilgrosso: Maybe not dangerous, but useless.
Please remove it, thanks.

ilgrosso:
@mmoayyed it seems all IT runs are failing with same error: https://travis-ci.org/github/apache/syncope/jobs/732715336#L744 - could you please take a look?

ilgrosso (approved):

mmoayyed:
Sure. Reviewing now.

PR #221, opened 2020-10-19
SYNCOPE-1593: signing/encryption algs per WA SAML2 SP

mmoayyed:
Allow SAML2 SPs in WA to have the ability to present their own requirements for signing algos, or blocking algos, etc. (i.e. an SP might require SHA-1 as the signing algorithm, or may want to block it)

Ref: https://issues.apache.org/jira/browse/SYNCOPE-1593


core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/auth/SAML2SP.java
```diff
@@ -73,4 +75,28 @@
     void setServiceProviderNameIdQualifier(String location);
 
     String getServiceProviderNameIdQualifier();
+
```
ilgrosso, line 78: I wonder if we could use an existing Enum or create new one to use instead of plain strings.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/auth/JPASAML2SP.java
```diff
@@ -192,4 +241,62 @@ public void setServiceProviderNameIdQualifier(final String serviceProviderNameId
...
+    public void setSigningSignatureBlackListedAlgorithms(final List<String> signingSignatureBlackListedAlgorithms) {
+        this.signingSignatureBlackListedAlgorithms = signingSignatureBlackListedAlgorithms;
+    }
+
+    @Override
+    public List<String> getEncryptionBlackListedAlgorithms() {
+        return encryptionBlackListedAlgorithms;
+    }
+
+    public void setEncryptionBlackListedAlgorithms(final List<String> encryptionBlackListedAlgorithms) {
```
ilgrosso, line 299: Please fix this code scanning alert.

ilgrosso (changes requested):

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/auth/SAML2SP.java
mmoayyed: Looks around. Can't find an existing enum, but it would be feasible to create an enum of all algorithms listed here:
https://www.w3.org/TR/xmlsec-algorithms/

Apache has a component that lists everything as a String constant:
https://santuario.apache.org/Java/api/org/apache/xml/security/encryption/XMLCipher.html

Perhaps this can be used in the enum we'd create? This seems like it would be consistent with the approach we take elsewhere in WA.

ilgrosso: I agree: FYI the given class is available at https://github.com/apache/santuario-xml-security-java/blob/master/src/main/java/org/apache/xml/security/encryption/XMLCipher.java

ilgrosso (approved):

PR #223, opened 2020-10-27
SYNCOPE-1595: Improve WA support for client app themes

mmoayyed:
This pull request,

- Adds support for `theme` to be defined on a per client-app basis
- Removes duplicate code from data-binders
- Fixes configuration of `wa-starter` module to add missing properties
- Cleans up Syncope's UI changes to rely on themes rather than overlaying html files

https://issues.apache.org/jira/browse/SYNCOPE-1595


core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/ClientAppDataBinderImpl.java
```diff
@@ -226,96 +172,59 @@ private void doUpdate(final OIDCRP clientApp, final OIDCRPTO clientAppTO) {
...
         return clientAppTO;
     }
 
     private void doUpdate(final CASSP clientApp, final CASSPTO clientAppTO) {
+        doUpdateCommon(clientApp, clientAppTO);
+
+        clientApp.setServiceId(clientAppTO.getServiceId());
+    }
+
+    private static CASSPTO getClientAppTO(final CASSP clientApp) {
```
ilgrosso, line 200: @mmoayyed could you please have a look at this?

ilgrosso (changes requested):

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/ClientAppDataBinderImpl.java
ilgrosso: I mean this: https://github.com/apache/syncope/pull/223/checks?check_run_id=1314109065

ilgrosso (approved):

PR #229, opened 2020-11-29
Upgrade WA to use CAS v6.3.0-RC5

mmoayyed:
Also fixes a number of minor issues regarding WebAuthN JSON serialization components.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -449,7 +449,7 @@ under the License.
 
     <pac4j.version>4.2.0</pac4j.version>
 
-    <cas.version>6.3.0-RC4</cas.version>
+    <cas.version>6.3.0-RC5</cas.version>
     <cas-client.version>3.6.2</cas-client.version>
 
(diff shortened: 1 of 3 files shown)


mmoayyed:
This looks to be ready.

Changes are:

- Had to add commons-codec dependency, since it was transitively pulled in by the saml2-idp module and provided via its "java-support" dependency. It's no longer made available since java-support was updated. Without this, pac4j-saml configuration fails to define idp/sp metadata for WA.

- Similar issues with okio. The okio dependency is pulled in by the duo module, which is really old and messes up the spring-cloud-config dependency.

ilgrosso:
@mmoayyed the changes look generally good to me: only, I'd either add `okio` to root `pom.xml` with version and remove version from `wa/starter/pom.xml`, or move both just to `wa/starter/pom.xml`

ilgrosso (approved):

PR #231, opened 2021-01-04
NOJIRA: skip copying legal resource files with test profile

mmoayyed:
Disables the copying of apidocs and legal-files if the test profile is activated. Seems to speed up the build to some degree.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2611,6 +2611,7 @@ under the License.
             </goals>
             <configuration>
               <outputDirectory>${project.build.outputDirectory}/META-INF</outputDirectory>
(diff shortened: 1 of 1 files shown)


ilgrosso:
The  `skipTests` profile is not actually a "test profile", it is rather a way to get full build without running any actual test.
Naturally, not recommended for any check or even release process, but still useful.

For these reasons, I don't think it is indeed a good idea to skip copying legal files, as by doing so the produced artifacts will not be the same as with other profiles.

mmoayyed:
Makes sense. Thanks for the update.

PR #233, opened 2021-01-11
[SYNCOPE-1609] Changes jsonb queries in order to improve performances

fmartelli:
Improve queries in order to avoid table joins

### DIFF
--- a/core/persistence-jpa-json/src/main/java/org/apache/syncope/core/persistence/jpa/dao/PGJPAJSONAnySearchDAO.java
+++ b/core/persistence-jpa-json/src/main/java/org/apache/syncope/core/persistence/jpa/dao/PGJPAJSONAnySearchDAO.java
@@ -18,9 +18,6 @@
  */
 package org.apache.syncope.core.persistence.jpa.dao;
 
-import static org.apache.syncope.core.persistence.jpa.dao.AbstractAnySearchDAO.RELATIONSHIP_FIELDS;
-import static org.apache.syncope.core.persistence.jpa.dao.AbstractDAO.LOG;
-
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -68,7 +65,7 @@ import org.apache.syncope.core.provisioning.api.utils.RealmUtils;
 
 public class PGJPAJSONAnySearchDAO extends AbstractJPAJSONAnySearchDAO {
 
-    protected static final String EMPTY_QUERY = "1=2";
+    protected static final String ALWAYS_FALSE_ASSERTION = "1=2";
 
     @Override
     protected void parseOrderByForPlainSchema(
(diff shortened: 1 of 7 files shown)


core/persistence-jpa-json/src/main/resources/pgjsonb/views.xml
```diff
@@ -47,13 +47,13 @@ under the License.
   </entry>
 
   <!-- user -->
-  <entry key="user_search">
+<!--  <entry key="user_search">
```
ilgrosso, line 50: Why comment this entry? If not needed anymore, please just remove.

core/persistence-jpa-json/src/main/java/org/apache/syncope/core/persistence/jpa/dao/PGJPAJSONAnySearchDAO.java
```diff
@@ -18,76 +18,57 @@
  */
 package org.apache.syncope.core.persistence.jpa.dao;
 
+import static org.apache.syncope.core.persistence.jpa.dao.AbstractAnySearchDAO.RELATIONSHIP_FIELDS;
```
ilgrosso, line 21: These two static imports shall be removed

core/persistence-jpa-json/src/main/java/org/apache/syncope/core/persistence/jpa/dao/PGJPAJSONAnySearchDAO.java
```diff
@@ -18,76 +18,57 @@
...
-                }
-
-                where.append(')');
-            } else {
-                where.append(',').append(searchView.name);
-            }
-            where.append(' ').append(searchView.alias);
-        });
-    }
+    protected static final String EMPTY_QUERY = "1=2";
```
ilgrosso, line 71: There is already a constant defined in superclass (`JPAAnySearchDAO`) with same name, please rename.

core/persistence-jpa-json/src/main/java/org/apache/syncope/core/persistence/jpa/dao/PGJPAJSONAnySearchDAO.java
```diff
@@ -18,76 +18,57 @@
...
-import org.apache.syncope.core.provisioning.api.serialization.POJOHelper;
-import org.apache.syncope.core.persistence.api.entity.JSONPlainAttr;
+import org.apache.syncope.core.persistence.api.entity.Realm;
 import org.apache.syncope.core.provisioning.api.utils.FormatUtils;
+import org.apache.syncope.core.provisioning.api.utils.RealmUtils;
 
 public class PGJPAJSONAnySearchDAO extends AbstractJPAJSONAnySearchDAO {
 
-    @Override
-    protected void processOBS(
```
ilgrosso, line 43: What about this method? Is it going to be removed completely? There might be subclasses with override: what should be done in such cases?

ilgrosso (changes requested):

ilgrosso:
@fmartelli would it be possible to:

1. open an improvement on JIRA describing the changes you want to apply
2. restore ordering of methods in `PGJPAJSONAnySearchDAO` so that it's easier to compare with current version?

Thanks!

core/persistence-jpa-json/src/main/java/org/apache/syncope/core/persistence/jpa/dao/PGJPAJSONAnySearchDAO.java
fmartelli: I didn't understand the question. 
PGJPAJSONAnySearchDAO does not need processOBS anymore while other classes need it.
Could you articulate this point?

fmartelli:
I've just opened the improvement 
https://issues.apache.org/jira/browse/SYNCOPE-1609

core/persistence-jpa-json/src/main/java/org/apache/syncope/core/persistence/jpa/dao/PGJPAJSONAnySearchDAO.java
ilgrosso: Nope, it is fine now, thanks!

ilgrosso (approved):
Great work, thanks! :+1:

PR #234, opened 2021-01-19
[SYNCOPE-1610] Refactoring ReconQuery to work with FIQL

ilgrosso:


mdisabatino (approved):

ilgrosso:
I am going to:

1. revert the changes on Console
2. change `ConnObjectTO` to bear, alongside with attributes, the FIQL query to uniquely identify the given object on the given External Resource; the idea is that such FIQL will generally be something like `__UID__==value` unless a `PushPolicy` with `PushCorrelationRule` is defined: in such cases it will be responsibility of the `PushCorrelationRule` implementation to generate the FIQL
3. change Console to use the FIQL from a given  `ConnObjectTO` instance to implement single pull

PR #237, opened 2021-01-26
[SYNCOPE-1613] Use latestExec startDate for "start" field of scheduled tasks

DmitriyBrashevets:


ilgrosso:
Thank you @DmitriyBrashevets

PR #239, opened 2021-01-27
reconciliation status: accept moreAttrsToGet also when search by anyKey

andrea-patricelli:


core/logic/src/main/java/org/apache/syncope/core/logic/ReconciliationLogic.java
```diff
@@ -237,7 +241,8 @@ public ReconStatus status(final String anyTypeKey, final String resourceKey, fin
         status.setOnSyncope(getOnSyncope(any, connObjectKeyItem, provision));
 
         List<ConnectorObject> connObjs = outboundMatcher.match(
-                connFactory.getConnector(provision.getResource()), any, provision, Optional.empty());
+                connFactory.getConnector(provision.getResource()), any, provision,
+                Optional.ofNullable(moreAttrsToGet.toArray(new String[] {})));
```
ilgrosso, line 245: You should use `Optional.of()` here instead of `Optional.ofNullable()`

ilgrosso (changes requested):

core/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/ReconciliationServiceImpl.java
```diff
@@ -98,7 +98,8 @@ public ReconStatus status(final ReconQuery query) {
         validate(query);
 
         if (query.getAnyKey() != null) {
-            return logic.status(query.getAnyTypeKey(), query.getResourceKey(), query.getAnyKey());
+            return logic.status(query.getAnyTypeKey(), query.getResourceKey(), query.getAnyKey(),
+                    query.getMoreAttrsToGet());
```
ilgrosso, line 102: `query,getMoreAttrsToGet()` can be `null` (see elsewhere methods in this class)

ilgrosso: You can for instance replace

```query.getMoreAttrsToGet()```

with

```java
Optional.ofNullable(query.getMoreAttrsToGet()).orElse(new HashSet<>())
```

andrea-patricelli: You're right, my bad. I'm going to fix

PR #240, opened 2021-01-28
Move SAML2 IdP settings to core packages

mmoayyed:


mmoayyed:
The next snapshot should be published in approx. 45mins. I will try to restart the CI process here again to make sure everything still passes.

PR #241, opened 2021-01-28
[SYNCOPE-1435] CSS modification for Console

DimaAy:


client/idrepo/console/src/main/java/org/apache/syncope/client/console/panels/DomainWizardBuilder.java
```diff
@@ -103,7 +103,7 @@ public Storage(final Domain domain) {
             add(new AjaxTextFieldPanel(
                     "jdbcURL",
                     "jdbcURL",
-                    new PropertyModel<>(domain, "jdbcURL")).setRequired(true));
+                    new PropertyModel<>(domain, "jdbcURL")).setRequired(true).addRequiredLabel());
```
ilgrosso, line 106: ```.setRequired(true).addRequiredLabel()``` should be replaced by ```.addRequiredLabel()``` as the latter's implementation contains the former call

ilgrosso: for all occurrences of course

ilgrosso (changes requested):

client/idm/console/src/main/resources/org/apache/syncope/client/console/wizards/resources/AbstractConnConfPanel.html
```diff
@@ -21,7 +21,7 @@
     <div id="emptyPlaceholder"/>
     <span wicket:id="connectorPropertiesContainer">
       <div class="pull-right">
-        <a style="position: fixed; top: 65px; right:60px;" wicket:id="check"  href="#">
+        <a style="position: fixed; top: 95px; right:420px;" wicket:id="check"  href="#">
```
mat-ale, line 24: Here it would be better to use `position: absolute; top: 0; right: 20px;`, because the rule with `fixed` seems to be not responsive when changing device or just resizing the window.

mat-ale: ...and a `z-index: 1` would be good to have in order not to hide the icon behind other fields!

client/idrepo/common-ui/src/main/resources/META-INF/resources/ui-commons/css/search.scss
```diff
@@ -36,25 +35,44 @@
     float: left;
     padding: 0 3px 0px 0px;
     display: inline-block !important;
+    .form-check{
+      margin-bottom: -50px;
+    }
+    .toggle.btn-xs {
```
mat-ale, line 41: as an improvement, for this kind of rules the related `.scss` style is:

```scss
.toggle {
  &.btn-xs {/*...*/}
}
```
and for this specific case I would do:
```scss
.toggle {
  &.btn {/*...*/}
}
```
instead, in order to avoid overriding a rule related to the default css grid classes (`xs`, `md` etc...).

client/idrepo/common-ui/src/main/resources/META-INF/resources/ui-commons/css/syncopeUI.scss
```diff
@@ -382,6 +399,16 @@ section#notifications .modal-lg {
   display: block;
 }
 
+.details-footer {
```
mat-ale, line 402: I think that:
```css
.details-footer {
   margin: 30px 0px 0px 0px;
   border: 1px solid #EEE;
   color: #888;
}
```
should be enough here

client/idrepo/common-ui/src/main/resources/META-INF/resources/ui-commons/css/syncopeUI.scss
```diff
@@ -665,6 +702,36 @@ div.listview-actions a {
...
+  width: 50%;
+  float: left;
+}
+
+.right.page-header{
+  padding-bottom: 9px;
+  border-bottom: 1px solid #dee2e6;
+  width: 50%;
+  float: right;
+}
```
mat-ale, line 728: Selectors with the same content could be appended with `,` e.g.:
```css
.left.page-header,
.right.page-header {
  padding-bottom: 9px;
  border-bottom: 1px solid #dee2e6;
  width: 50%;
}

.left.page-header{
  float: left;
}
.right.page-header{
  float: right;
}
```

client/idm/console/src/main/resources/org/apache/syncope/client/console/wizards/resources/AbstractConnConfPanel.html
mat-ale: after a deeper check this solution is not good, since the `Override?` flag bothers the check icon.
A totally better solution is to edit that html file as following:
```html
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:wicket="http://wicket.apache.org">
  <wicket:panel>
    <div id="emptyPlaceholder"/>
    <span wicket:id="connectorPropertiesContainer">
      <a class="conn-check-wrapper" wicket:id="check"  href="#">
        <i class="fa fa-heartbeat fa-2x"></i>
      </a>
<!-- ... -->
```

and the `syncopeUi.scss` file needs to include:
```scss
.scrollable-tab-content {
  padding-top: 30px;
  /* other already existing rules */
}
.conn-check-wrapper {
  position: absolute;
  top: 0;
  right: 25px;
  z-index: 1;
}
```

ilgrosso (approved):
LGTM

mat-ale:
LGTM too, good job!

mat-ale (approved):

DimaAy:
Thank you @ilgrosso and @mat-ale

PR #244, opened 2021-03-08
Upgrade to CAS 6.4 RC2 and pac4j v5 RC2

mmoayyed:


ext/saml2sp4ui/logic/src/main/java/org/apache/syncope/core/logic/SAML2SP4UILogic.java
```diff
@@ -252,10 +254,10 @@ public SAML2Request createLoginRequest(
                 RequestedAuthnContext requestedAuthnContext = requestedAuthnContextProvider.get();
                 SAML2Configuration cfg = saml2Client.getConfiguration();
```
ilgrosso, line 255: It seems this line is not needed anymore (see CodeQL scannin)

wa/starter/src/main/resources/application.properties
```diff
@@ -40,9 +40,14 @@ management.endpoint.loggers.enabled=true
...
 
-cas.events.core.enabled=false
+# Cache service definitions for 5 minutes
+cas.service-registry.cache.duration=PT5M
+
+# Reload services and hydrate the cache every 5 minutes
+cas.service-registry.schedule.repeat-interval=PT5M
+cas.service-registry.schedule.start-delay=PT30S
 
-server.servlet.session.tracking-modes=cookie
```
ilgrosso, line 45: Any reason to remove `server.servlet.session.tracking-modes=cookie` ?

ilgrosso (changes requested):
LGTM, just a couple of minor issues.

Also, please double check, when modifying a `.properties` file, that all variants in the whole source tree are also updated accordingly, thanks.

wa/starter/src/main/resources/application.properties
mmoayyed: It appears that this line is unnecessary and a duplicate in fact. Here's the first reference:
https://github.com/apache/syncope/blob/master/wa/starter/src/main/resources/application.properties#L31

ilgrosso: Nice catch :-)

ilgrosso (approved):
LGTM

PR #247, opened 2021-03-16
Auto-configure session tracking modes

mmoayyed:
When running with initializr in separate servlet container, session tracking modes are left to their defaults and `SessionConfiguringInitializer` backs away, as it only affects embedded servlet containers. This makes both COOKIE and URL modes available to the underlying response, which cause very strange behavior with URL encoding and inclusion of `jsessionid` in the URL, which in turn causes Spring Security to block the request as it sees `;` as a malicious character.

In this patch, the behavior for `SessionConfiguringInitializer` is partially emulated to set the session tracking modes based on configured application properties.


ilgrosso (approved):
LGTM

PR #258, opened 2021-04-06
Switch to CAS 6.4

mmoayyed:


ilgrosso:
@mmoayyed please merge latest changes from `master` to fix build with Payara.

About CAS / Pac4j upgrade, I am getting the same exceptions locally: https://github.com/apache/syncope/runs/2275643884?check_suite_focus=true#step:7:4202

mmoayyed:
It seems like the exceptions are due to the fact that xmlsectool ships with and leaks opensaml 4.0.1, whereas the configuration requires 4.1.0. I will play around with some exclusion rules to see what I can do. Alternative, I might be able to force the opensaml version to use 4.1.0, which I think is likely the better option.

mmoayyed:
Fixed the exceptions locally. Further testing however shows that there are number of other issues regarding logout operations for SAML2 and OIDC. So, at the moment it's not feasible to upgrade to RC3, and I will work on fixes to make sure we can upgrade to RC4 in a few weeks. 

Changing PR to Draft status.

fit/wa-reference/src/test/java/org/apache/syncope/fit/ui/SAML2SP4UIITCase.java
```diff
@@ -156,6 +158,20 @@ public static void idpSetup() {
...
+    public void fetchSpMetadata() throws Exception {
+        try (CloseableHttpClient httpclient = HttpClients.createDefault()) {
+            HttpClientContext context = HttpClientContext.create();
+            context.setCookieStore(new BasicCookieStore());
+
+            HttpGet get = new HttpGet(WA_ADDRESS + "/sp/metadata");
+            CloseableHttpResponse response = httpclient.execute(get, context);
+            assertEquals(HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+            String responseBody = EntityUtils.toString(response.getEntity());
+            assertNotNull(responseBody);
```
ilgrosso, line 171: I'd rather change this with

```
assertFalse(responseBody.isEmpty());
```

ilgrosso (approved):
LGTM

PR #262, opened 2021-04-22
[SYNCOPE-1630] (1) centralizing security checks in DAOs and (2) impro…

ilgrosso:
…ve JPAAnySearchDAO


ilgrosso:
Still TODO: add a pre-defined `GROUP_OWNER` Role where to place the set of Entitlements to assign, rather than do it by code

PR #264, opened 2021-05-20
[SYNCOPE-1635] Use singleton bean for each domain for each rule with conf

DmitriyBrashevets:
…conf

### DIFF
--- a/core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
+++ b/core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
@@ -18,11 +18,11 @@
  */
 package org.apache.syncope.core.spring;
 
-import groovy.lang.GroovyClassLoader;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import groovy.lang.GroovyClassLoader;
 import org.apache.syncope.common.lib.policy.AccountRuleConf;
 import org.apache.syncope.common.lib.policy.PasswordRuleConf;
 import org.apache.syncope.common.lib.policy.PullCorrelationRuleConf;
@@ -31,14 +31,16 @@ import org.apache.syncope.common.lib.report.ReportletConf;
 import org.apache.syncope.core.persistence.api.ImplementationLookup;
 import org.apache.syncope.core.persistence.api.dao.AccountRule;
 import org.apache.syncope.core.persistence.api.dao.PasswordRule;
-import org.apache.syncope.core.persistence.api.dao.Reportlet;
-import org.apache.syncope.core.persistence.api.entity.Implementation;
 import org.apache.syncope.core.persistence.api.dao.PullCorrelationRule;
 import org.apache.syncope.core.persistence.api.dao.PushCorrelationRule;
+import org.apache.syncope.core.persistence.api.dao.Reportlet;
+import org.apache.syncope.core.persistence.api.entity.Implementation;
 import org.apache.syncope.core.provisioning.api.serialization.POJOHelper;
+import org.apache.syncope.core.spring.security.AuthContextUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.support.AbstractBeanDefinition;
+import org.springframework.beans.factory.support.DefaultListableBeanFactory;
 
 public final class ImplementationManager {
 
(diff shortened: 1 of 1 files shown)


core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -72,7 +73,7 @@
         }
     }
 
-    public static Optional<AccountRule> buildAccountRule(final Implementation impl)
+    public static Optional<AccountRule> buildAccountRule(final String domain, final Implementation impl)
```
ilgrosso, line 76: same as above

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -96,7 +97,7 @@
         }
     }
 
-    public static Optional<PasswordRule> buildPasswordRule(final Implementation impl)
+    public static Optional<PasswordRule> buildPasswordRule(final String domain, final Implementation impl)
```
ilgrosso, line 100: same as above

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -120,7 +121,7 @@
         }
     }
 
-    public static Optional<PullCorrelationRule> buildPullCorrelationRule(final Implementation impl)
+    public static Optional<PullCorrelationRule> buildPullCorrelationRule(final String domain, final Implementation impl)
```
ilgrosso, line 124: same as above

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -145,7 +146,7 @@
         }
     }
 
-    public static Optional<PushCorrelationRule> buildPushCorrelationRule(final Implementation impl)
+    public static Optional<PushCorrelationRule> buildPushCorrelationRule(final String domain, final Implementation impl)
```
ilgrosso, line 149: same as above

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -216,28 +217,27 @@
...
-                try {
-                    bean = (T) ApplicationContextProvider.getBeanFactory().
-                            createBean(clazz, AbstractBeanDefinition.AUTOWIRE_BY_TYPE, false);
-                    ApplicationContextProvider.getBeanFactory().registerSingleton(clazz.getName(), bean);
-                } catch (IllegalStateException e) {
-                    LOG.debug("While attempting to register {}", clazz.getName(), e);
+            String domainableBeanNameWithConf = domain + clazz.getName();
+            DefaultListableBeanFactory beanFactory = ApplicationContextProvider.getBeanFactory();
+
+            if (beanFactory.containsSingleton(domainableBeanNameWithConf)) {
```
ilgrosso, line 225: The same exact `if` with statement is also performed in the `synchronized` block: why?

DmitriyBrashevets: This is a double-check lock pattern. Thread takes monitor and executes the logic regarding initialization. Another thread waits until the the monitor is realeased. As soon as first thread releases monitor -> second threads proceeds its work and reaches the additional `if` clause. As the bean already exists (created by first thread) the second thread is not creating a bean again and not registers it as a singleton.


Do you see any disadvantages it this?

ilgrosso: I see, there is some literature abut double-check lock pattern, and some is not very pleasant with it, especially for Java implementations.

What would be the disadvantage of simply removing the first `if` and leaving the whole logic in the `synchronized` block?

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
```diff
@@ -48,7 +49,7 @@
 
     private static final Map<String, Class<?>> CLASS_CACHE = Collections.synchronizedMap(new HashMap<>());
 
-    public static Optional<Reportlet> buildReportlet(final Implementation impl)
+    public static Optional<Reportlet> buildReportlet(final String domain, final Implementation impl)
```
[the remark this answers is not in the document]
DmitriyBrashevets: I thought that it's becoming more understandable (visible) if public methods have the additional parameter declared explicitly. But can be removed...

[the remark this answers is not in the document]
DmitriyBrashevets: removed addtional param

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
DmitriyBrashevets: removed addtional param

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
DmitriyBrashevets: removed addtional param

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
DmitriyBrashevets: removed addtional param

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
DmitriyBrashevets: removed addtional param

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
DmitriyBrashevets: Ah... I see... Maybe java optimization will remove this logic from bytecode when the method will become hot.

ilgrosso (approved):

core/spring/src/main/java/org/apache/syncope/core/spring/ImplementationManager.java
ilgrosso: I am not completely convinced, but let that be :-)

PR #266, opened 2021-05-25
Remove unused PasswordGenerator class in the installer

coheigea:
It's not used since the move to use SSHA-256 by default in MavenUtils.


ilgrosso (approved):
LGTM - I suppose this is relation with 39d1e61798e

coheigea:
Yep exactly.

PR #268, opened 2021-05-26
Disable CXF Services Listing

coheigea:
As Syncope 2.1.x is stuck on CXF 3.2.x, it is vulnerable to:

http://cxf.apache.org/security-advisories.data/CVE-2020-13954.txt.asc

I confirmed with this PR you can no longer see the services page (http://localhost:9080/syncope/rest/services)


ilgrosso:
@coheigea when you disable service listing, does doc page still works, a.k.a. /syncope/ which output is similar to http://syncope.apache.org/rest/2.1/index.html ?

PR #273, opened 2021-06-14
Switch to CAS 6.4 RC6

mmoayyed:


core/am/logic/src/main/java/org/apache/syncope/core/logic/wa/GoogleMfaAuthAccountLogic.java
```diff
@@ -57,6 +58,26 @@ public void deleteFor(final String owner) {
         });
     }
 
+    @PreAuthorize("hasRole('" + IdRepoEntitlement.ANONYMOUS + "')")
+    public void deleteFor(final long id) {
```
ilgrosso, line 62: Please rename this method as `delete()`

pom.xml
```diff
@@ -2045,40 +2005,6 @@ under the License.
         <enabled>true</enabled>
       </releases>
     </repository>
-
-    <!-- Disable some Spring repositories pulled in by dependencies - see
```
ilgrosso, line 2049: Why removing disabled Spring repositories here?

ilgrosso (changes requested):

ilgrosso:
@mmoayyed shouldn't pac4j be upgraded to 5.1.0 as well?

pom.xml
mmoayyed: Why would you want to keep disabled repositories?

ilgrosso: Isn't the information provided in the given link still valid?

mmoayyed: It is, but disabling a repository is effectively the same as a no-op. It's dead code, and certainly, the repositories are no longer needed and have been removed from upstream references. If you're telling maven, "don't use this repository for anything", then it would make sense to remove "unused" code.

ilgrosso: There are dependencies that have such repos defined hence Syncope builds will report permission errors; we added such definitions to stop pulling from such repos.

mmoayyed: Could you point out what those dependencies are? Can we upgrade those? And I don't follow. If the repository is 100% disabled, how are those dependencies resolved?

ilgrosso: When you remove such disabled repo definitions, you will notice that some 3rd party dependencies will be attempted to be downloaded from Spring repositories, failing with permission errors.
Anyway, let's keep the repo removal, I am going to build from your branch to see if I can still see such errors.

mmoayyed:
Good point on pac4j; thanks for the note. I upgraded it.

mmoayyed:
Update: I am working on fixing/diagnosing issues with OIDC/SAML2 tests. There are changes that need to be mad to handle dynamic issuers for OIDC, and almost none belong in the WA. The WA fixes are mostly sanitization of configuration, rather than any major changes, and I am hoping to finalize by early next week.

wa/starter/src/main/resources/log4j2.xml
```diff
@@ -37,9 +37,14 @@ under the License.
 
   <loggers>
 
-    <asyncLogger name="org.apereo.cas" additivity="false" level="INFO">
+    <asyncLogger name="org.apereo.cas" additivity="false" level="DEBUG">
```
ilgrosso, line 40: I think this `DEBUG` and the one following should be reverted to `INFO` - this file is supposed to be used by deployments.

ilgrosso (changes requested):
@mmoayyed it seems to be working again now, good job!
Besides the small remark about logging level in one file, are you willing to merge this PR now or would you prefer to wait until next RC?

wa/starter/src/main/resources/log4j2.xml
mmoayyed: No objections, but wouldn't you say that having DEBUG level on is incredibly useful for a deployment as a starting point to troubleshoot and figure out what's going on? That is [the very first thing](https://apereo.github.io/cas/development/installation/Troubleshooting-Guide.html#review-logs) we do for deployment, or ask people to do when they trying to diagnose issues.

mmoayyed: OTOH, I think we should revert. You're right; I would prefer the system to be production ready OOTB.

ilgrosso (changes requested):

mmoayyed:
I would prefer to wait since the next RC would be out on a few days. I will probably do another merge with master right before the rc release to make sure everything runs ok and then we can merge this after the rc is out.

ilgrosso (approved):

PR #275, opened 2021-06-25
[SYNCOPE-1435] New wicket enduser

loredicola:


ilgrosso (changes requested):
Please fix all license headers.

ilgrosso (approved):

PR #279, opened 2021-07-19
code cleanup; diamond ops && lambda refs

mmoayyed:
This is purely a code clean-up patch:

- Replace explicit types with their equivalent diamond operator
- Replace methods calls with method references


ilgrosso:
@mmoayyed thanks for taking care of such cleanup: would it be possible to remove the formatting changes from the PR?

ilgrosso:
@mmoayyed it seems that Travis CI is failing due to a JDK bug like https://stackoverflow.com/questions/54775253/jdk-11-0-2-compilation-fails-with-javac-npe-on-anonymous-parameterized-class-typ
Could you please have a look?

mmoayyed:
I don't think formatting changes can be undone, since the fix are not manually made and are a result of automated tools. If you wish to see the changes without formatting changes, you can use this link: https://github.com/apache/syncope/pull/279/files?w=1

Travis CI issues are the result of an old bug in JDK that has since been patches; nothing we can do without significant changes to the Travis CI which should be disabled anyway, since that is not the main CI and just creates extra unnecessary work.

ilgrosso:
@mmoayyed I see about formatting changes.

About Travis CI, instead, the checks done there are currently not performed with GitHub Actions (as they require Docker and might involve some work); hence, we cannot ATM disable Travis CI.

mmoayyed:
Github Actions natively supports Docker with almost no effort; so probably a better task would be to disable Travis CI and instead focus the effort to move the Travis CI tests over to GH actions, rather than maintaining two separate builds. I can certainly help with that.

ilgrosso:
@mmoayyed please see https://github.com/apache/syncope/pull/281

ilgrosso:
@mmoayyed please merge latest changes from `master` to get the updated GH actions setup

ilgrosso (approved):

PR #281, opened 2021-07-21
Attempting to ditch completely Travis CI for GH Actions

ilgrosso:


.github/workflows/fit_Elasticsearch.yml
```diff
@@ -0,0 +1,57 @@
...
+  schedule:
+    - cron: '0 13 * * 4'
+
+jobs:
+  fit_Elasticsearch:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
```
mmoayyed, line 34: You can likely take this out. You're not testing against various combinations in a matrix. It'a always Java and JDK 11.

mmoayyed (commented):
Looks great! Just a comment on the use of matrix.  Did you mean to include other JDK versions for tests? If not, then matrix can be taken out from the config.

PR #284, opened 2021-07-28
Clean up: Simplify optional call chains

mmoayyed:
- Simplify `if(Optional.isPresent())` that can be rewritten in the functional style.
- Simplify optional call chains to use `isPresent()` or `isEmpty()` correctly.

A more accurate description of changes can be seen at: https://github.com/apache/syncope/pull/284/files?w=1


ilgrosso (approved):
LGTM thanks for taking care of such code improvements @mmoayyed !

PR #286, opened 2021-08-13
[SYNCOPE-1640] fixed removal of empty enumeration key on multivalue p…

andrea-patricelli:
…aram and enum schema definition

### DIFF
--- a/client/console/src/main/java/org/apache/syncope/client/console/commons/PropertyList.java
+++ b/client/console/src/main/java/org/apache/syncope/client/console/commons/PropertyList.java
@@ -174,13 +174,11 @@ public abstract class PropertyList<T> implements List<String> {
         final StringBuilder builder = new StringBuilder();
         int index = 0;
         for (String str : enumerationValues) {
-            if (str != null) {
-                if (index > 0) {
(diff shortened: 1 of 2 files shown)


client/console/src/main/java/org/apache/syncope/client/console/commons/PropertyList.java
```diff
@@ -174,13 +174,11 @@ public static String getEnumValuesAsString(final List<String> enumerationValues)
         final StringBuilder builder = new StringBuilder();
```
ilgrosso, line 174: Why not replacing the whole method with

```
return enumerationValues.stream().
                map(v -> StringUtils.isBlank(v) ? StringUtils.EMPTY : v.trim()).
                collect(Collectors.joining(SyncopeConstants.ENUM_VALUES_SEPARATOR));
```

?

ilgrosso (changes requested):

client/console/src/main/java/org/apache/syncope/client/console/commons/PropertyList.java
andrea-patricelli: Sure, LGTM! It should work seamlessly.

ilgrosso (approved):

PR #289, opened 2021-09-17
[SYNCOPE-1643] Chaning of user realm doesn't trigger provisioning

DmitriyBrashevets:
Signed-off-by: Dmitriy Brashevets <dmitriy.brashevets@united-security-providers.ch>

### DIFF
--- a/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
+++ b/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
@@ -384,9 +384,6 @@ public class UserDataBinderImpl extends AbstractAnyDataBinder implements UserDat
 
         AnyUtils anyUtils = anyUtilsFactory.getInstance(AnyTypeKind.USER);
 
-        // realm
-        setRealm(user, userPatch);
-
         // password
         String password = null;
         boolean changePwd = false;
@@ -413,6 +410,9 @@ public class UserDataBinderImpl extends AbstractAnyDataBinder implements UserDat
         // Save projection on Resources (before update)
         Map<String, ConnObjectTO> beforeOnResources =
                 onResources(user, userDAO.findAllResourceKeys(user.getKey()), password, changePwd);
+
+        // realm
+        setRealm(user, userPatch);
 
         // username
         if (userPatch.getUsername() != null && StringUtils.isNotBlank(userPatch.getUsername().getValue())) {


core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -675,6 +675,14 @@ private boolean isPasswordMapped(final ExternalResource resource) {
                             collect(Collectors.toSet()));
         }
 
+        if (userPatch.getRealm() != null) {
```
ilgrosso, line 678: please check that the patch is effectively changing the current Realm

ilgrosso (changes requested):

ilgrosso:
@DmitriyBrashevets please apply similar changes to `GroupDataBinderImpl` and `AnyObjectDataBinderImpl`

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -384,9 +384,6 @@ private boolean isPasswordMapped(final ExternalResource resource) {
 
         AnyUtils anyUtils = anyUtilsFactory.getInstance(AnyTypeKind.USER);
 
-        // realm
```
ilgrosso, line 387: why moving the `setRealm()` method invocation?

ilgrosso (changes requested):
I was meaning to place the change originally proposed into the `AbstractDataBinder#setRealm` method body.

DmitriyBrashevets:
> @DmitriyBrashevets please apply similar changes to `GroupDataBinderImpl` and `AnyObjectDataBinderImpl`

For groups the provisioning works well when the realm is changed.

If the org.apache.syncope.core.provisioning.java.data.AbstractAnyDataBinder#setRealm is executed after the before-projection is created then `org.apache.syncope.core.provisioning.java.data.AbstractAnyDataBinder#propByRes` calculates resources correctly as there may be changes in the `org.apache.syncope.common.lib.to.AttrTO` with schema "__NAME__" because of connObjectLink.
This leads to a consistent behavior for UserDataBinder and GroupDataBinder.

So, I'd say that moving of the execution of `org.apache.syncope.core.provisioning.java.data.AbstractAnyDataBinder#setRealm` below the creation of projections solves the issue. Please, look at: https://github.com/apache/syncope/blob/ad2523680b68e843cba35261c350521875f81945/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/GroupDataBinderImpl.java#L203
Please, tell:
1. if `org.apache.syncope.core.provisioning.java.data.AbstractAnyDataBinder.setRealm(Any<?>, AnyPatch)` still needs to be modfied
2. if you're fine with moving of `org.apache.syncope.core.provisioning.java.data.AbstractAnyDataBinder#setRealm` below

ilgrosso:
@DmitriyBrashevets I understand that:

1. only moving the `setRealm()` invocation, with no further changes, fixes `UserDataBinderImpl`'s behavior
2. no issues with `GroupDataBInderImpl`

If these are correct, then please just apply the same change as (1) into `AnyObjectDataBinderImpl` and we should be all set, thanks.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
DmitriyBrashevets: I described in https://github.com/apache/syncope/pull/289#issuecomment-921725709

DmitriyBrashevets:
> @DmitriyBrashevets I understand that:
> 
> 1. only moving the `setRealm()` invocation, with no further changes, fixes `UserDataBinderImpl`'s behavior
> 2. no issues with `GroupDataBInderImpl`
> 
> If these are correct, then please just apply the same change as (1) into `AnyObjectDataBinderImpl` and we should be all set, thanks.

The `AnyObjectDataBinderImpl`  doesn't require any changes the `setRealm()` is executed after the creation of the projection

PR #290, opened 2021-09-17
NOJIRA - Turn off bean proxying for WA; fixes typo

mmoayyed:
This pull request turns off and disables CGLIB bean proxying for the WA module. This is the recommended practice from Spring which would help WA in the future to compile better and easier against GraalVM. Without bean proxying, bean methods can no longer call each other directly during construction, as there is no proxy to intercept. Instead, references either need to be passed as auto-wired parameters (the approach used here), or they need to be injected into configuration classes using injections. (The latter approach is generally more preferable, if the number of injected parameters grows unwieldy) 

Also fixes a bean name typo (missing `e`).

This approach might also be used for #287 as well.


ilgrosso (approved):

ilgrosso:
Thanks for providing these very useful background information: without knowing exactly the context, I am already taking this approach in #287 , maybe not everywhere yet.

PR #296, opened 2021-11-08
Switch CAS to use 6.5.x / Spring Boot 2.6.x

mmoayyed:
- Upgrade to Spring Boot 2.6.x and Spring Cloud 3.1.x
- Upgrade to CAS 6.5 SNAPSHOT (until RC4 is released by end of December 2021)


client/idrepo/console/pom.xml
```diff
@@ -124,7 +124,7 @@ under the License.
     </dependency>
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-slf4j18-impl</artifactId>
+      <artifactId>log4j-slf4j-impl</artifactId>
```
ilgrosso, line 127: Why this (and similar changes) replacing `log4j-slf4j18-impl` with its "general JDK" counterpart?

pom.xml
```diff
@@ -430,7 +430,7 @@ under the License.
 
     <camel.version>3.13.0</camel.version>
 
-    <slf4j.version>2.0.0-alpha5</slf4j.version>
+    <slf4j.version>1.7.32</slf4j.version>
```
ilgrosso, line 433: Are we switching back here to 1.7.32 in order to align with CAS, correct?

mmoayyed: Yes. Staying on alpha seems to cause a few issues:

1. Seems like Boot 2.6 RC1 has issues recognizing this given logger binding have changed; possibly because boot's dependency management of slf4j has changed.
2. The alpha version is quite unstable, does not work against graalvm and native-images with the spring-native project. This is the main reason for CAS to fallback on 1.7.x.

If preferable, I can try to only override this for CAS. I see that slf4j has been on alpha for a long time, and there is no official/public plan as to when this might get released. Log4j is also waiting for slf4j to stabilize before the build compatible bindings for it. So what works now, may not work in the future, and is in fact sort of luckily works.

ilgrosso: Not a big deal, let's stick with `1.7.32` everywhere, thanks.

client/idrepo/console/pom.xml
mmoayyed: (Copying the answer here as well so it's available for the public)

> This is to mainly to align with CAS, and to also align with the now-stable slf4j version, to avoid these:

```
java.lang.ClassCastException: class org.slf4j.helpers.NOPLoggerFactory cannot be cast to class 
org.apache.logging.slf4j.Log4jLoggerFactory (org.slf4j.helpers.NOPLoggerFactory and 
org.apache.logging.slf4j.Log4jLoggerFactory are in unnamed module of loader 
org.springframework.boot.loader.LaunchedURLClassLoader
```

Ref: https://github.com/apereo/cas/pull/5261
Ref: https://github.com/apereo/cas/pull/5264

ilgrosso: Thanks for explanation, LGTM!

ilgrosso (changes requested):
The PR looks good to me.

I think it is also good to update `spring-cloud-contract-wiremock` to `${spring-cloud-gateway.version}` by restoring https://github.com/apache/syncope/blob/master/pom.xml#L1891-L1893

Also: do you have idea why GH is displaying changes from other commits into this PR's changes?

mmoayyed:
Sure. I updated the PR.

I think GH might be confused after the branch was merged with master. This branch was sync'd with master once, but it seems like GH does not recognize the changes as those that might already be merged with master.

ilgrosso (approved):

PR #299, opened 2021-12-20
Remove bean proxies from configuration components

mmoayyed:
This pull request begins to remove support for bean proxying and to disable cglib proxies in general for Syncope components. The default will then switch to native JDK proxies instead, in attempts to assist with native build.


mmoayyed:
Will re-open later when CI changes have settled.

ilgrosso:
@mmoayyed TIL logs are under `fit/*-reference/target/log` - not `logs`

PR #301, opened 2022-01-03
Remove CGLib proxies from Spring configuration components - part 1

mmoayyed:
This pull request starts to remove CGLib proxies from Configuration classes. The changes are quite large, but they for the most part tend to be cosmetic. Here is a summary:

1. `proxyBeanMethods` is now set to false for `@Configuration` annotations
2. As a result, bean methods can no longer call each other directly, since there is no proxy to intercept. Therefor, intra-bean calls have been refactored to use parameter injection.
3. Where possible, `@Component` annotations are removed and the component creation has been replaced with a `@Bean` instead. Furthermore, field injections that contain auto-wired references are also removed from such components, and are replaced with c-tor injection at creation time.
4. `@Autowired` annotations are removed from `@Bean` methods; `@Resource` annotations are also replaced with `@Qualifier`. 
5. In order to assist with PRs running CI builds on forks, CI builds can now also trigger for PRs as long as the PR branch name begins with `pr-`
6. CAS build is now switched to use RC4.

This is only the first part of this effort; CGLib proxies are still created for many `@Bean`s and `@Component` references, in scenarios where the bean does not have a parent interface, or is created with a concrete type. This would need to change, which also impacts how `@ConditionalOnClass` elements are used.


.github/workflows/codeql-analysis.yml
```diff
@@ -18,7 +18,7 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [master,2_1_X]
+    branches: [master,2_1_X,pr-*]
```
ilgrosso, line 21: I don't understand such changes to GH workflow: has anything changed lately? Otherwise, such changes are not required to let the workflows run on PRs.

fit/wa-reference/src/main/resources/log4j2.xml
```diff
@@ -63,7 +63,11 @@ under the License.
       <appender-ref ref="main"/>
     </asyncLogger>
 
-    <asyncLogger name="org.apache.cxf" additivity="false" level="ERROR">
+    <asyncLogger name="org.apache.cxf" additivity="false" level="DEBUG">
+      <appender-ref ref="main"/>
+    </asyncLogger>
+
+    <asyncLogger name="org.springframework.security" additivity="false" level="DEBUG">
```
ilgrosso, line 70: Please revert, as this file is taken by archetype as reference for new projects.

core/starter/src/main/java/org/apache/syncope/core/starter/SyncopeCoreApplication.java
```diff
@@ -105,4 +221,31 @@ public SyncopeCoreStart keymasterStart(final DomainHolder domainHolder) {
     public KeymasterStop keymasterStop() {
         return new KeymasterStop(NetworkService.Type.CORE);
     }
+
```
ilgrosso, line 224: Is introducing these two new interfaces / default implementations the only way to go?

.github/workflows/codeql-analysis.yml
mmoayyed: When you are workin on a fork of the codebase, and push changes to a branch other than master, CI jobs will not run on your fork for that branch unless this change is in place. PR rules are separate from this and are not affected.

core/starter/src/main/java/org/apache/syncope/core/starter/SyncopeCoreApplication.java
mmoayyed: Yes, and this approach should be applied to all other bean implementations that do not conform or implement an interface. Specially for event-listeners in the absence of proxies. The `SyncopeCoreInfoContributor` does some very strange things today and is combined to also react to a listener and collect data; This violates many other principals , but I did not change the way it works; mainly the event listener config is moved out of the contributor.

mmoayyed: For example,

```
    @ConditionalOnMissingBean
    @Bean
    public AuthDataAccessor authDataAccessor() {}
```

Will create a CGLib proxy; since the class is a concrete type and there is no interface to use for the proxy object. The general approach is to turn `AuthDataAccessor` into an interface, with a default implementation class, or something to that effect.

.github/workflows/codeql-analysis.yml
ilgrosso: Ok, I see, thanks.

core/starter/src/main/java/org/apache/syncope/core/starter/SyncopeCoreApplication.java
ilgrosso: I understand the problem now: I'd say then we need to handle this whole item apart for next PR (keeping the current changes in this PR, anyway) - we're talking here about "residual" beans, as several ones were converted into the interface + default impl approach in the recent past.

I'd say that such new interfaces / default impls should be treated properly in separate files, too, rather than being embedded in Application / Configuration classes.

mmoayyed: Yes, I definitely agree. That would be a fairly comprehensive change, likely to impact many modules and I plan to tackle that separately via a series of PRs. That makes it easier to test incrementally as well, and the only *sure* way I know to test is to look at the beans actuator endpoint and make sure no reference to "CGLIB" is found in the output. It's fairly slow, but certain :)

ilgrosso (approved):

PR #315, opened 2022-02-15
Upgrading CAS to 6.5.0, moving wa-reference as https

mmoayyed:


mmoayyed:
Now seeing a lot of:

```
12:03:11.754 ERROR org.apache.syncope.common.keymaster.client.api.ServiceOps - Could not unregister NetworkService{type=SRA, address=http://localhost:8080/}
java.util.concurrent.CompletionException: javax.ws.rs.ProcessingException: javax.net.ssl.SSLHandshakeException: SSLHandshakeException invoking https://localhost:9443/syncope/rest/keymaster/networkServices?action=unregister: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314) ~[?:?]
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319) [?:?]
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1702) [?:?]
	at java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1692) [?:?]
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290) [?:?]
	at java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020) [?:?]
	at java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656) [?:?]
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594) [?:?]
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183) [?:?]
Caused by: javax.ws.rs.ProcessingException: javax.net.ssl.SSLHandshakeException: SSLHandshakeException invoking https://localhost:9443/syncope/rest/keymaster/networkServices?action=unregister: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

Testing with `mvn test -Dtest=CASSRAITCase`.

pom.xml
```diff
@@ -446,9 +446,9 @@ under the License.
 
     <modernizer-maven.version>2.3.0</modernizer-maven.version>
 
-    <pac4j.version>5.2.1</pac4j.version>
+    <pac4j.version>5.3.1</pac4j.version>
 
-    <cas.version>6.5.0-RC4</cas.version>
+    <cas.version>6.5.1-SNAPSHOT</cas.version>
```
mmoayyed, line 451: Note that this version bump is necessary because of:
https://github.com/spring-cloud/spring-cloud-config/issues/1645 

This will get better in 6.6.x

ilgrosso:
If possible, we should address the following items as well:

* continuous statements appearing in the logs as
```
WARN  org.springframework.web.context.support.SpringBeanAutowiringSupport - Current WebApplicationContext is not available for processing of SyncopeWAGoogleMfaAuthTokenRepository: Make sure this class gets constructed in a Spring web application after the Spring WebApplicationContext has been initialized. Proceeding without injection.
```
* properties managed by [WAConfigService](https://github.com/apache/syncope/blob/master/common/am/rest-api/src/main/java/org/apache/syncope/common/rest/api/service/wa/WAConfigService.java]) do not seem to be actually loaded on push

mmoayyed:
A few updates:

- There are no references to `SpringBeanAutowiringSupport` in CAS 6.5; This should be removed and should not show up; however, the warning is entirely benign and can be turned off via logging. 
- Do you happen to know which properties are not loaded? All or only a select few?

ilgrosso:
> Do you happen to know which properties are not loaded? All or only a select few?

It starts empty, but when I add something like

```
cas.authn.attribute-repository.syncope.url=https://localhost:9443/syncope/rest
cas.authn.attribute-repository.syncope.basic-auth-username=admin
cas.authn.attribute-repository.syncope.basic-auth-password=XXXX
cas.authn.attribute-repository.syncope.search-filter= username=={user}
```

I can see it that attribute resolution is not working, while adding the same props in `fit/wa-reference/src/main/resources/wa-embedded.properties` all is working as expected.

mmoayyed:
I see; thanks for the details.

This seems simple-enough to fix; unfortunately like most other seemingly-simple problems, it's quite complicated to work it out. Mainly because of this:
https://github.com/spring-cloud/spring-cloud-config/issues/1645

In summary, this is because attribute resolution registered by Syncope is only conditionally activated via:

```java
@ConditionalOnProperty("cas.authn.attribute-repository.syncope.url")
public class SyncopePersonDirectoryConfiguration { }
```

This means, when you start up empty, the configuration class and all the other beans in it will get ignored and skipped. Then, when you refresh by adding a value (such as URL), Spring application context will fail to refresh beans, because there is nothing to refresh to begin with. Refresh requests and scope only work in scenarios in the application context hierarchy where there is a reference to a bean that can be refreshed; beans or configuration classes that were skipped during the startup and application context initialization will never be refreshable, because they are not re-created. IOW, refresh requests only work best when there is a setting whose existing value changes from A to B; if there was no A to begin with, or if A is being removed, refresh requests may fall short. 

So it's very difficult to: keep things refreshable, and, make sure the context is not polluted with *noise* (i.e. beans that exists and do nothing).

For now, we might have to accept noise; that is to remove `@ConditionalOnProperty("cas.authn.attribute-repository.syncope.url")` and let the context create empty beans, only to be refreshed later. This might be a good compromise for now in the 6.5.1 line.

This is being worked on in 6.6 so that conditional beans on props continue to work correctly, though as you can imagine, this goes quite deep and will take a bit of time to get right.

ilgrosso:
Ok, your explanation makes definitely sense to me.

So, we should consider (and maybe document) as manageable by `WAConfigService` not all the available CAS properties, but only the ones that are not used to define conditions on Spring beans; those ones shall be instead placed in some `wa-*.properties` file.

mmoayyed:
Yes I agree. That's good thing to do at least for the time being while this stuff gets worked out, with a small note to revisit this later in the future upgrades.

ilgrosso:
`CASSRAITCase#web` is failing. The reason seems to be that accessing

http://localhost:8080/protected/get?key1=value1&key2=value2&key2=value3&key3=an%20url%20encoded%20value%3A%20this%21

(served by SRA) triggers redirect to WA

https://localhost:9443/syncope-wa/login?service=https%3A%2F%2Flocalhost%3A8080%2Fprotected%2Fget%3Fkey1%3Dvalue1%26key2%3Dvalue2%26key2%3Dvalue3%26key3%3Dan%2Burl%2Bencoded%2Bvalue%253A%2Bthis%2521

which throws 

```
17:57:07.559 ERROR org.apereo.cas.services.web.support.RegisteredServiceResponseHeadersEnforcementFilter - Service unauthorized
org.apereo.cas.services.UnauthorizedServiceException: Service unauthorized
        at org.apereo.cas.services.RegisteredServiceAccessStrategyAuditableEnforcer.lambda$execute$6(RegisteredServiceAccessStrategyAuditableEnforcer.java:194) ~[cas-server-core-services-api-6.5.1-SNAPSHOT.jar:6.5.1-SNAPSHOT]
        at java.util.Optional.orElseGet(Optional.java:369) ~[?:?]
        at org.apereo.cas.services.RegisteredServiceAccessStrategyAuditableEnforcer.execute(RegisteredServiceAccessStrategyAuditableEnforcer.java:192) ~[cas-server-core-services-api-6.5.1-SNAPSHOT.jar:6.5.1-SNAPSHOT]
```

Visiting https://localhost:9443/syncope-wa/actuator/registeredServices returns (among others)

```
{
      "@class": "org.apereo.cas.services.RegexRegisteredService",
      "serviceId": "http://localhost:8080/.*",
      "name": "org.apache.syncope.fit.sra.CASSRAITCase",
      "id": 4,
      "attributeReleasePolicy": {
        "@class": "org.apereo.cas.services.ReturnMappedAttributeReleasePolicy",
        "allowedAttributes": {
          "@class": "java.util.TreeMap",
          "syncopeUserAttr_email": "email",
          "syncopeUserAttr_firstname": "given_name",
          "syncopeUserAttr_fullname": "name",
          "syncopeUserAttr_surname": "family_name",
          "syncopeUserMemberships": "groups"
        }
      },
      "authenticationPolicy": {
        "@class": "org.apereo.cas.services.DefaultRegisteredServiceAuthenticationPolicy",
        "requiredAuthenticationHandlers": [
          "java.util.HashSet",
          [
            "DefaultSyncopeAuthModule"
          ]
        ]
      }
    }
```

that looks actually fine.

@mmoayyed can you please have a look?

Side note: any reason why the `registered-services` actuator endpoint now seems to require `Content-Type: application/json` header? I remember it used to work directly from browser in the past.

mmoayyed:
The reason this is failing is because:

- Your service is registered as `http://localhost:8080/.*`
- The service parameter sent to CAS is `https://localhost:8080/protected/get?....`

As you see the pattern does not match. The pattern should allow `https`; it's missing an `s`.

ilgrosso (approved):

PR #319, opened 2022-03-02
[SYNCOPE-1666] added security answer hashing

andrea-patricelli:


core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/user/User.java
```diff
@@ -54,8 +55,14 @@
...
     String getSecurityAnswer();
+    
+    String getClearSecurityAnswer();
 
-    void setSecurityAnswer(String securityAnswer);
+    void setEncodedSecurityAnswer(String securityAnswer);
+    
+    void setSecurityAnswer(String securityAnswer,  CipherAlgorithm cipherAlgoritm);
+
+    boolean canDecodeSecurityAnswer();
```
ilgrosso, line 65: Is this method really needed? Isn't just a copy of `canDecodePassword()`?
We might unify the two methods in one `canDecodeSecrets()`

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/user/User.java
```diff
@@ -54,8 +55,14 @@
     void setSecurityQuestion(SecurityQuestion securityQuestion);
 
     String getSecurityAnswer();
+    
+    String getClearSecurityAnswer();
 
-    void setSecurityAnswer(String securityAnswer);
+    void setEncodedSecurityAnswer(String securityAnswer);
+    
+    void setSecurityAnswer(String securityAnswer,  CipherAlgorithm cipherAlgoritm);
```
ilgrosso, line 63: This method would allow to set  a different `cipherAlgorithm` for the user than the one set for password.
The risk is that `user#getCipherAlgorith` can return the wrong value when used to check password or security question values via `Encryptor#verify`.

I think we should change `User` interface to:

1. remove `cipherAlgorithm` param from `setPassword()` and `setSecurityAnswer()` methods
1. add a separate `setCipherAlgorithm()` method, failing if `cipherAlgorithm` is already set, unless the provided parameter is `null`
1. have `setPassword()` and `setSecurityAnswer()` to internally set `cipherAlgorithm` to the value from conf

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -681,32 +711,21 @@ private boolean isPasswordMapped(final ExternalResource resource) {
...
-            accountTO.getPlainAttrs().add(new AttrTO.Builder().
-                    schema(plainAttr.getSchema().getKey()).
-                    values(plainAttr.getValuesAsStrings()).build());
-        });
-
-        accountTO.getPrivileges().addAll(account.getPrivileges().stream().
-                map(Entity::getKey).collect(Collectors.toList()));
-
-        return accountTO;
+        return getLinkedAccountTO(account, true);
```
ilgrosso, line 714: Is there any call for `getLinkedAccountTO(account, false)`?

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -756,20 +775,21 @@ public UserTO getUserTO(final User user, final boolean details) {
             // memberships
             userTO.getMemberships().addAll(
                     user.getMemberships().stream().map(membership -> getMembershipTO(
-                    user.getPlainAttrs(membership),
-                    derAttrHandler.getValues(user, membership),
-                    virAttrHandler.getValues(user, membership),
-                    membership)).collect(Collectors.toList()));
+                            user.getPlainAttrs(membership),
```
ilgrosso, line 778: please revert this formatting-only change

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -756,20 +775,21 @@ public UserTO getUserTO(final User user, final boolean details) {
...
+                            derAttrHandler.getValues(user, membership),
+                            virAttrHandler.getValues(user, membership),
+                            membership)).collect(Collectors.toList()));
 
             // dynamic memberships
             userTO.getDynMemberships().addAll(
                     userDAO.findDynGroups(user.getKey()).stream().map(group -> new MembershipTO.Builder().
-                    group(group.getKey(), group.getName()).
-                    build()).collect(Collectors.toList()));
+                            group(group.getKey(), group.getName()).
```
ilgrosso, line 786: please revert this formatting-only change

ext/flowable/logic/src/main/java/org/apache/syncope/core/logic/UserRequestLogic.java
```diff
@@ -126,7 +126,7 @@ public UserRequest startRequest(
     protected void securityChecks(final String username, final String entitlement, final String errorMessage) {
         if (!AuthContextUtils.getUsername().equals(username)
                 && !AuthContextUtils.getAuthorities().stream().
-                        anyMatch(auth -> entitlement.equals(auth.getAuthority()))) {
+                anyMatch(auth -> entitlement.equals(auth.getAuthority()))) {
```
ilgrosso, line 129: please revert this formatting-only change

ilgrosso (changes requested):

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/user/User.java
andrea-patricelli: Yes, totally agree. In fact I didn't add `setCipherAlgorithm()` from `setSecurityAnswer()`implementation.
Going to perform your proposed changes

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
andrea-patricelli: ATM no.

andrea-patricelli: yes, if the param `return.password.value` is false

andrea-patricelli: here is possible https://github.com/apache/syncope/blob/c8d86c28ee95192e6f81eb67f7fb89f3f6f72a07/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java#L790-L792

ilgrosso: Sounds good, thanks.

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/user/User.java
andrea-patricelli: Sure. That's much better

core/persistence-api/src/main/java/org/apache/syncope/core/persistence/api/entity/user/User.java
andrea-patricelli: do we also remove from `setEncodedPassword()`? Seems the same case

ilgrosso: `setEncodedPassword` is used internally by DAO, it is required

andrea-patricelli: I've notice that disallowing setting cipher algorithm once set could impact cases like the one in this test
https://github.com/apache/syncope/blob/2_1_X/core/provisioning-java/src/test/java/org/apache/syncope/core/provisioning/java/MappingManagerImplTest.java#L61-L132

Could it be too much restrictive? That is not so usual situation, but could lead to errors if for some reason an admin decides to change password hashing algorithm, with all its consequences...

ilgrosso: Tests can't be a reason to change code / design.
Just user `ReflectionTestUtils#setField` if nothing else works.

andrea-patricelli: Ok, yes, I'll use that. My wonder was to cases when we want to change password cipher algo for some reason. In my experience this happened only once.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/user/JPALinkedAccount.java
```diff
@@ -157,10 +166,9 @@ public void setEncodedPassword(final String password, final CipherAlgorithm ciph
     }
 
     @Override
-    public void setPassword(final String password, final CipherAlgorithm cipherAlgoritm) {
+    public void setPassword(final String password) {
         try {
-            this.password = ENCRYPTOR.encode(password, cipherAlgoritm);
-            this.cipherAlgorithm = cipherAlgoritm;
+            this.password = ENCRYPTOR.encode(password, this.cipherAlgorithm);
```
ilgrosso, line 171: What happens if `this.cipherAlgorithm` is `null`? it should take the default value from conf IIRC

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/user/JPAUser.java
```diff
@@ -241,21 +244,20 @@ public void removeClearPassword() {
...
 
     @Override
-    public void setPassword(final String password, final CipherAlgorithm cipherAlgoritm) {
+    public void setPassword(final String password) {
         this.clearPassword = password;
 
         try {
-            this.password = ENCRYPTOR.encode(password, cipherAlgoritm);
-            this.cipherAlgorithm = cipherAlgoritm;
+            this.password = ENCRYPTOR.encode(password, cipherAlgorithm);
```
ilgrosso, line 260: What happens if `this.cipherAlgorithm` is `null`? it should take the default value from conf IIRC

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/user/JPAUser.java
```diff
@@ -424,9 +435,28 @@ public String getSecurityAnswer() {
...
+
+        this.securityAnswer = securityAnswer;
+    }
+
     @Override
     public void setSecurityAnswer(final String securityAnswer) {
         this.securityAnswer = securityAnswer;
+
+        try {
+            this.securityAnswer = ENCRYPTOR.encode(securityAnswer, cipherAlgorithm);
```
ilgrosso, line 455: What happens if `this.cipherAlgorithm` is `null`? it should take the default value from conf IIRC

core/persistence-jpa/src/test/java/org/apache/syncope/core/persistence/jpa/inner/UserTest.java
```diff
@@ -292,7 +303,45 @@ public void testPasswordGenerator() {
         assertNotNull(password);
 
         User user = userDAO.find("c9b2dec2-00a7-4855-97c0-d854842b4b24");
-        user.setPassword(password, CipherAlgorithm.SHA);
+        user.setPassword(password);
         userDAO.save(user);
     }
+
+    @Test
+    public void testPasswordGeneratorFailing() {
```
ilgrosso, line 311: Please avoid naming test methods as `test...`

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/MappingManagerImpl.java
```diff
@@ -479,6 +479,7 @@ private String getIntValue(final Realm realm, final Item orgUnitItem) {
 
     protected String decodePassword(final Account account) {
         try {
+            LOG.info("Decode password {} in {}", account.getPassword(), account.getCipherAlgorithm());
```
ilgrosso, line 482: Avoid logging secrets, even if encoded

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -144,17 +135,42 @@ public UserTO getAuthenticatedUserTO() {
...
     }
 
+    private void setCipherAlgorithm(final Account account) {
+        if (account.getCipherAlgorithm() == null) {
+            account.setCipherAlgorithm(
+                    CipherAlgorithm.valueOf(confDAO.find("password.cipher.algorithm", CipherAlgorithm.AES.name())));
+        }
+    }
+
+    private RuntimeException aggregateException(
```
ilgrosso, line 164: how many methods are calling this new one? If it's just one, remove the new method and place its statements in the caller

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -755,21 +783,20 @@ public UserTO getUserTO(final User user, final boolean details) {
 
             // memberships
             userTO.getMemberships().addAll(
-                    user.getMemberships().stream().map(membership -> getMembershipTO(
-                    user.getPlainAttrs(membership),
-                    derAttrHandler.getValues(user, membership),
-                    virAttrHandler.getValues(user, membership),
-                    membership)).collect(Collectors.toList()));
+                    user.getMemberships().stream().map(membership -> getMembershipTO(user.getPlainAttrs(membership),
```
ilgrosso, line 786: Please revert this formatting-only change

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
```diff
@@ -755,21 +783,20 @@ public UserTO getUserTO(final User user, final boolean details) {
...
+                            derAttrHandler.getValues(user, membership),
+                            virAttrHandler.getValues(user, membership),
+                            membership)).collect(Collectors.toList()));
 
             // dynamic memberships
             userTO.getDynMemberships().addAll(
                     userDAO.findDynGroups(user.getKey()).stream().map(group -> new MembershipTO.Builder().
-                    group(group.getKey(), group.getName()).
-                    build()).collect(Collectors.toList()));
+                            group(group.getKey(), group.getName()).build()).collect(Collectors.toList()));
```
ilgrosso, line 794: Please revert this formatting-only change

ilgrosso (changes requested):

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/data/UserDataBinderImpl.java
andrea-patricelli: I would keep it since it's used by `setPassword` and `setSecurityAnswer`.

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/user/JPALinkedAccount.java
ilgrosso: We already do: https://github.com/apache/syncope/blob/2_1_X/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/entity/user/JPAUser.java#L194

ilgrosso (approved):

ilgrosso:
LGTM @andrea-patricelli please go ahead squash and merge if tests are succeeding

PR #320, opened 2022-03-03
[SYNCOPE-1652] Add missing conf items

ilgrosso:


mmoayyed (approved):
LGTM.

One thing to bear in mind in the future perhaps; an application can be assigned multiple access strategies in a chained manner and combined using AND/OR rules. Something to keep in mind and possibly plan for in the future when the time is right.

PR #329, opened 2022-03-21
Fix for classpath scan

andrea-patricelli:


client/idrepo/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
```diff
@@ -263,9 +263,10 @@ public void load() {
 
     @SuppressWarnings("unchecked")
     public <T> List<Class<? extends T>> getClasses(final Class<T> reference) {
-        return classes.getOrDefault(reference.getName(), List.of()).stream().
-                map(clazz -> (Class<? extends T>) clazz).
-                collect(Collectors.toList());
+        return classes.entrySet().stream()
```
ilgrosso, line 266: What is the rationale behind this change? Can you provide an example where the original code fails and this works, instead?

ilgrosso (changes requested):

client/idrepo/console/src/main/java/org/apache/syncope/client/console/init/ClassPathScanImplementationLookup.java
ilgrosso: Your samples above are mixing Core and Console classes.

PR #336, opened 2022-04-06
[SYNCOPE-1672] Can't retrieve Java classes when add java implementation for Password Rule, Account Rule and Reportlet on Console

SamuelGaro:


ilgrosso (approved):
LGTM thanks @SamuelGaro 

We'll probably need to review the whole `ImplementationInfoProvider` mechanism in the future, but for the moment this fix is needed.

PR #349, opened 2022-06-01
SYNCOPE-1681: Support LDAP backends for Google Authenticator MFA

mmoayyed:
See https://issues.apache.org/jira/browse/SYNCOPE-1681

This pull request re-enables the `refresh` endpoint, and allows google authenticator to be managed via an LDAP backend that is controlled by WA. When LDAP settings are defined, WA backs away in favor of the new LDAP backend. 

Note that once LDAP is configured, removing the settings from WA (to disable LDAP and re-enable Syncope) will not work. To disable LDAP, one must restart WA. This is a current limitation of Spring Cloud; removed/blanked settings are not accounted for in refresh events. 

This is based on CAS 6.5.5-SNAPSHOT, which fixes a bug with configuration binding and refresh events; to be released this Friday. 

Also removed a few unused imports here and there.


ilgrosso (approved):

mmoayyed:
Sounds good, sure. We can tackle that later.

PR #355, opened 2022-06-22
[SYNCOPE-1682] Introducing AttrRepo entity for AttrReleasePolicy

ilgrosso:


ilgrosso:
The changes here require https://github.com/apereo/cas/pull/5482 to work properly, at least for LDAP, JDBC and Syncope attribute repositories

PR #360, opened 2022-07-13
[SYNCOPE-1689] Consolidate Provision, Mapping and Items into single JSON column

ilgrosso:
The purpose of this PR is to explore the possibility to consolidate a set of JPA entities from their own tables into single JSON column.

The idea is that such strategy should improve the general provisioning performance, especially during propagation, as no joins will be needed to fetch the required information.

At least some comparison tests shall be performed against current `master` branch and this PR to understand if the proposed change is effective or not.

It could be an idea to do the same with Connector and Resource capabilities.


ilgrosso:
Very interesting results from some smoke performance tests.

I've been using [ilgrosso/syncopeRestClient](https://github.com/ilgrosso/syncopeRestClient) to create users with propagation against Syncope instances started via `mvn -Pdebug,all` under `fit/core-reference` directory.

| | `master` | this PR |
|--------------|-----------|------------|
| 1000 users | 1145426 ms (~19 minutes)  | 189939 ms (~3 minutes) |

PR #373, opened 2022-09-02
Support multi-arch docker builds

mmoayyed:
In particular, produce images compatible with ARM architectures.

### DIFF
--- a/docker/console/pom.xml
+++ b/docker/console/pom.xml
@@ -141,6 +141,12 @@ under the License.
               <name>apache/syncope-console:%v</name>
               <build>
                 <dockerFileDir>${project.build.outputDirectory}</dockerFileDir>
+                <buildx>
+                  <platforms>
+                    <platform>linux/amd64</platform>
+                    <platform>linux/arm64</platform>
+                  </platforms>
+                </buildx>
               </build>
             </image>
           </images>
--- a/docker/console/src/main/resources/Dockerfile
+++ b/docker/console/src/main/resources/Dockerfile
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-FROM eclipse-temurin:11-focal
+FROM eclipse-temurin:17-focal
 MAINTAINER dev@syncope.apache.org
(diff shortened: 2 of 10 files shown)


common/am/rest-api/src/main/java/org/apache/syncope/common/rest/api/service/OIDCJWKSService.java
```diff
@@ -18,6 +18,8 @@
  */
 package org.apache.syncope.common.rest.api.service;
 
+import io.swagger.v3.oas.annotations.Parameter;
```
ilgrosso, line 21: Please avoid adding unused imports, as they will trigger checkstyle failures.

ilgrosso (changes requested):
Thank you @mmoayyed 
Since there are no actual GH actions building Docker images, I think we should check locally if the proposed changes are working, then merge.

Maybe, since we are at it, I would like to propose to move the base image from JDK 11 to 17.

BTW any chance you consider to create PRs from a fresh or rebased from an updated `master` branch? :sweat_smile:

mmoayyed:
Sure, I corrected the commit log. 

You can test the builds locally of course, but the true test would be to actually run the containers on different platforms. I can sort of confirm the arm build works correctly, but you likely have to try out the amd builds. 

Base images are now on jdk 17.

ilgrosso:
Thank you @mmoayyed I'll test the images on my `amd64`: there are a few [docker-compose samples](https://github.com/apache/syncope/tree/master/docker/src/main/resources/docker-compose) in case

ilgrosso:
@mmoayyed I can confirm that Docker images are built correctly and that I am able to run them on my `amd64`.

mmoayyed:
Great news. Should we merge, or would you prefer to wait for the build status to turn green first?

ilgrosso:
@mmoayyed it should be enough to merge from `master`, I've just pushed a commit to fix the build.
Anyway, no need to wait for it as GH actions don't check for Docker: please go ahead.

PR #379, opened 2022-10-07
Update CAS to 6.6.1

mmoayyed:


ilgrosso:
@mmoayyed I think you should include here the revert of 

https://github.com/apache/syncope/blob/master/wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/AuthModulePropertySourceMapper.java#L271-L275

to its former version, e.g.

```java
return prefix("cas.authn.mfa.simple.", CasCoreConfigurationUtils.asMap(props));
```

PR #381, opened 2022-10-12
SYNCOPE-1699: Extract key from path  if undefined in request body

mmoayyed:
Please see https://issues.apache.org/jira/browse/SYNCOPE-1699

### DIFF
--- a/core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/AbstractService.java
+++ b/core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/AbstractService.java
@@ -61,20 +61,18 @@ public abstract class AbstractService implements JAXRSService {
 
     protected String getActualKey(final AnyDAO<?> dao, final String pretendingKey) {
         String actualKey = pretendingKey;
-        if (uriInfo.getPathParameters(true).containsKey("key")) {
-            final String keyInPath = uriInfo.getPathParameters(true).get("key").get(0);
-            if (actualKey == null) {
-                actualKey = keyInPath;
-            } else if (!actualKey.equals(keyInPath)) {
-                SyncopeClientException sce = SyncopeClientException.build(ClientExceptionType.InvalidRequest);
-                sce.getElements().add("Key specified in request does not match key in the path");
-                throw sce;
-            }
-        }
         if (actualKey == null) {
             SyncopeClientException sce = SyncopeClientException.build(ClientExceptionType.InvalidRequest);
             sce.getElements().add("Key is null");
             throw sce;
+        }
+        if (uriInfo.getPathParameters(true).containsKey("key")) {
+            final String keyInPath = uriInfo.getPathParameters(true).get("key").get(0);
+            if (!actualKey.equals(keyInPath)) {
+                SyncopeClientException sce = SyncopeClientException.build(ClientExceptionType.InvalidRequest);
+                sce.getElements().add("Key specified in request does not match key in the path");
+                throw sce;
+            }
         }
         if (!SyncopeConstants.UUID_PATTERN.matcher(actualKey).matches()) {
             actualKey = dao.findKey(actualKey);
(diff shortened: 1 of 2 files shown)


core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/AbstractService.java
```diff
@@ -61,8 +61,23 @@ public abstract class AbstractService implements JAXRSService {
 
     protected String getActualKey(final AnyDAO<?> dao, final String pretendingKey) {
         String actualKey = pretendingKey;
-        if (!SyncopeConstants.UUID_PATTERN.matcher(pretendingKey).matches()) {
-            actualKey = dao.findKey(pretendingKey);
+        if (uriInfo.getPathParameters(true).containsKey("key")) {
```
ilgrosso, line 64: Can we add here `actualKey == null` as first condition the `if` clause? I'd rather avoid decoding path parameters if not strictly needed

ilgrosso (changes requested):

ilgrosso (changes requested):

core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/AbstractService.java
mmoayyed: You would, if the path parameter has a key that does not match the supplied key in the request body. You could in theory be passing one key in the URL, and use another key in the body.

ilgrosso: what I was meaning is that if you call `getActualKey(dao, null)` the current proposed implemenation will not even look at path parameters, but just raise the exception `Key is null`

ilgrosso (changes requested):

core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/AbstractService.java
ilgrosso: This is how I'd refactor the current proposal:

```java
    protected String getActualKey(final AnyDAO<?> dao, final String pretendingKey) {
        String actualKey = pretendingKey;
        if (actualKey == null && uriInfo.getPathParameters(true).containsKey("key")) {
            final String keyInPath = uriInfo.getPathParameters(true).get("key").get(0);
            if (actualKey != null && !actualKey.equals(keyInPath)) {
                SyncopeClientException sce = SyncopeClientException.build(ClientExceptionType.InvalidRequest);
                sce.getElements().add("Key specified in request does not match key in the path");
                throw sce;
            }
        }
        if (actualKey == null) {
            SyncopeClientException sce = SyncopeClientException.build(ClientExceptionType.InvalidRequest);
            sce.getElements().add("Key is null");
            throw sce;
        }
        if (!SyncopeConstants.UUID_PATTERN.matcher(actualKey).matches()) {
            actualKey = dao.findKey(actualKey);
        }

        return actualKey;
    }
 ```

Comment (ilgrosso) [reply core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/service/AbstractService.java:69]: you're right about the mismatch in my proposal; still, the current one will not even attempt to look at path parameters if the payload key is `null`

So it seems we are attempting to pursue to objectives here:
1. avoid NPE in case the payload bears a `null` key
2. if both path parameter and payload's  `key` are provided, check if they are the same

**Comment by ilgrosso** [comment]
I think you can also add a test case to `UserIssuesITCase` like the following, which would fail with NPE in the current `master`

```java
    @Test
    public void issueSYNCOPE1699() throws JsonProcessingException {
        UserTO userTO = createUser(UserITCase.getUniqueSample("syncope1337@apache.org")).getEntity();

        UserUR req = new UserUR();
        req.setUsername(new StringReplacePatchItem.Builder().value("newUsername" + getUUIDString()).build());

        WebClient webClient = WebClient.create(ADDRESS + "/users/" + userTO.getKey(), ADMIN_UNAME, ADMIN_PWD, null).
                accept(MediaType.APPLICATION_JSON_TYPE).
                type(MediaType.APPLICATION_JSON_TYPE);

        Response response = webClient.invoke("PATCH", JSON_MAPPER.writeValueAsString(req));
        assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
    }
```

mmoayyed:
I think my original implementation of this was correct. Please take another look. I included your test as well as one extra one to show how this would work. (Thanks for the test By the way!)

PR #402, opened 2023-01-17
SYNCOPE-1723 remove non-reproducible entries from git.properties

hboutemy:
see https://issues.apache.org/jira/browse/SYNCOPE-1723

this PR fixes one source of non-reproducible bits

for full path in properties, it would be easy to change value in pom.xml https://github.com/apache/syncope/blob/master/pom.xml#L513 to drop the absolute path, but I don't know if this will break for the builder: what i doubt is that people consuming the resulting artifact can use the value, because the path does not point to anything useful beyond the builder

### DIFF
--- a/archetype/pom.xml
+++ b/archetype/pom.xml
@@ -131,18 +131,18 @@ under the License.
         <includes>
           <include>core.properties</include>
         </includes>
-        <targetPath>${project.build.outputDirectory}/archetype-resources/core/src/main/resources</targetPath>
+        <targetPath>archetype-resources/core/src/main/resources</targetPath>
       </resource>
       <resource>
         <directory>../core/provisioning-java/src/main/resources</directory>
-        <targetPath>${project.build.outputDirectory}/archetype-resources/core/src/main/resources</targetPath>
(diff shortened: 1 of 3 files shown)


ilgrosso:
Thank for this PR @hboutemy 

I see no issues with the proposed change in this PR, e.g. excluding some GIT properties.

Abouth full-path filtered properties, we haven't been able so far to find viable alternatives without breaking the build process.
Please consider that (a) this hasn't been an issue so far and (b) that [all the provided ways to use Syncope artifacts](https://syncope.apache.org/docs/3.0/getting-started.html#obtain-apache-syncope) are using different filtering for such properties.

hboutemy:
@ilgrosso ok, this filtered properties issue is harder: let's see later
I added 2 other fixes
with the 3 fixes in place, the 7 non-reproducible artifact from releases 3.0.0 and 3.0.1 will drop to only 3 = these filtered properties

ilgrosso:
@hboutemy I have verified that your changes to archetype are not affecting any feature so I'll go ahead and merge this PR to `master` then cherry-pick to `3_0_X` (which is the actual branch from which 3.0.x releases are cut).

PR #405, opened 2023-01-19
[SYNCOPE-1723] Removing static paths from generated artifacts

ilgrosso:
Need to remove the following Maven properties:

- [x] conf.directory
- [x] connid.location
- [X] log.directory


ilgrosso:
@hboutemy this is the follow up from #402 

First results are encouraging, I'll try to see if all static paths can be removed altogether.

hboutemy:
wow, real great work: kudos

PR #408, opened 2023-02-07
[SYNCOPE-1728] fixes wrong label in BeanPanel and avoid not serializa…

andrea-patricelli:
…ble exception


ilgrosso:
@andrea-patricelli the fix looks fine but you still have a checkstyle violation reported.

Maybe nextime before pushing it could be an idea to run at least

```
mvn checkstyle:check
```

from the root directory.

andrea-patricelli:
> @andrea-patricelli the fix looks fine but you still have a checkstyle violation reported.
> 
> Maybe nextime before pushing it could be an idea to run at least
> 
> ```
> mvn checkstyle:check
> ```
> 
> from the root directory.

Yes, sure. My bad. Going to fix.

ilgrosso (approved):

PR #411, opened 2023-02-15
Console: Realm search bar enlargement and better positioning for small screens

loredicola:


ilgrosso (approved):
Thanks @loredicola nice improvement

PR #415, opened 2023-02-23
[Syncope 1734] ElasticSearch event is not created when external resource give back attribute mapped like uidOnCreate

SamuelGaro:


ilgrosso (approved):

ilgrosso:
Thank you @SamuelGaro

PR #419, opened 2023-03-02
Adds the possibility to specify a logo for a ClientApp

fmartelli:
This is mapped on the logo properties of a RegisteredService


common/am/lib/src/main/java/org/apache/syncope/common/lib/to/ClientAppTO.java
```diff
@@ -53,6 +53,8 @@ public abstract class ClientAppTO implements NamedEntityTO {
 
     private String description;
 
+    private String logo;
```
ilgrosso, line 56: Need to include this new property in `equals()` and `hashCode()` methods.

ilgrosso (changes requested):

ilgrosso (approved):

PR #422, opened 2023-03-08
[SYNCOPE-1739] Wrong volume mapping for source code in fit docker profile

SamuelGaro:


archetype/src/main/resources/archetype-resources/fit/pom.xml
```diff
@@ -300,6 +300,26 @@ under the License.
           </deployables>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <version>1.10</version>
```
ilgrosso, line 306: Please remove this `<version>` tag, version is taken from parent pom (and is currently set to `3.3.0`).

ilgrosso (changes requested):

ilgrosso (approved):

PR #455, opened 2023-05-03
Service PR (not for merge) to under GH action failure

ilgrosso:


ilgrosso:
I've possibly found the culprit: less frequent checks waiting for WA refresh + excluding `org.apereo.cas.configuration.model.core.CasServerProperties` from refresh by adding

```
spring.cloud.refresh.never-refreshable=com.zaxxer.hikari.HikariDataSource,org.apereo.cas.configuration.model.core.CasServerProperties
```

according to https://docs.spring.io/spring-cloud-commons/docs/3.1.6/reference/html/#refresh-scope

PR #472, opened 2023-06-13
SYNCOPE-1765: encrypt/decrypt values/settings for AuthModuleConf

mmoayyed:
cc @fmartelli 

https://issues.apache.org/jira/browse/SYNCOPE-1765

Allow AuthModuleConf implementations to support selective encryption/decryption of sensitive settings. For backward compatibility, values that cannot be decrypted are returned back verbatim.

### DIFF
--- a/common/am/lib/src/main/java/org/apache/syncope/common/lib/AbstractLDAPConf.java
+++ b/common/am/lib/src/main/java/org/apache/syncope/common/lib/AbstractLDAPConf.java
@@ -577,4 +577,9 @@ public abstract class AbstractLDAPConf implements Serializable {
     public void setBinaryAttributes(final List<String> binaryAttributes) {
         this.binaryAttributes = binaryAttributes;
     }
+
+    @Override
+    protected AbstractLDAPConf clone() throws CloneNotSupportedException {
+        return (AbstractLDAPConf) super.clone();
+    }
 }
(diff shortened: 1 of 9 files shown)


ilgrosso:
Why not including `AttrRepoConf` as well?

fmartelli:
@mmoayyed @ilgrosso 
actually I'd prefer something different, like as the approach you can find [here](https://apereo.github.io/cas/6.6.x/configuration/Configuration-Properties-Security-SpringCloud.html) for CAS.

It would be nice to have the possibility to specify encrypted values for whatever properties into authmodules, attrrepos and general WA properties, Maybe just specifying a prefix {cipher} like as we use to do with CAS.

The question: if we specify a property by providing the [encrypted value with {cas-cipher} prefix](https://apereo.github.io/cas/6.6.x/configuration/Configuration-Properties-Security-CAS.html), what will be interpreted by WA at the Apply configuration phase. Maybe we already have this feature for free ... just some minor reworks.

Please, let me know what you think about.
Thank you

mmoayyed:
What you should be able to do today is:

1. Find any setting that WA supports, and encrypt it using the same strategy that CAS allows
2. Go to the console, or any other configuration area/file in WA, and specify the setting value as `{cas-cipher}<ENCRYPTED_VALUE_HERE>`
3. The setting will be picked up by WA and decrypted accordingly.

fmartelli:
I think this could be enough.
I'll be glad to try out within today.

I'll let you know. Stay tuned

mmoayyed:
That does look correct, but I think the missing piece here is that WA bootstrapping has its own logic for loading properties, and those do not pass through the CAS cipher. This bit, we can add.

Note that `cas.standalone.configuration-security.psw` needs to be taught to WA at bootstrap time (that is when you launch WA) using a system property or environment variable.

fmartelli:
Hi @mmoayyed, I speciefied cas.standalone.configuration-security.psw into a spring profile so, yes, it should be loaded at bootstrap time.

Can you close this PR and open a new one in order to add the bit that is currently missing?

mmoayyed:
Sure. I'll probably get to this tomorrow. Thanks for the details!

PR #507, opened 2023-08-22
Typo fix

acarullo:
changed the value of an IT localized label


ilgrosso:
Thank you @acarullo nice spot!

PR #547, opened 2023-11-11
Attempting to understand why OpenSearch ITs keep failing on master

ilgrosso:
Not meant to be merged directly, just trying to understand why OpenSearch fail to start on master branch.


ilgrosso:
Ok so it seems the [startup issue](https://github.com/spring-projects/spring-boot/issues/38186) was fixed by https://github.com/spring-projects/spring-boot/commit/9eda564d063e347e7f937e61cd8c84819c670604 after Spring Boot 3.2.0-RC2 so it should be enough to just wait for next release and upgrade.

PR #440, opened 2023-04-12
[SYNCOPE-1748] Fix incorrectly named bulkMaxPayloadSize property

philipptrenz:
This PR resolves an naming issue, probably introduced by the merge of https://github.com/apache/syncope/pull/436, which causes an error within the web console of Syncope when opening the SCIM v2 extension panel:

```
Property could not be resolved for class: class org.apache.syncope.common.lib.scim.SCIMGeneralConf expression: bulkMaxMaxPayloadSize
WicketMessage: An error occurred while getting the model object for Component: [AjaxSpinner [Component id = spinner, page = org.apache.syncope.client.console.pages.SCIMConfPage, path = body:content:body:container:content:tabbedPanel:panel:bulkMaxMaxPayloadSize:spinner, type = com.googlecode.wicket.jquery.ui.form.spinner.AjaxSpinner, isVisible = true, isVersioned = false, markup = [markup = jar:file:/opt/syncope/lib/syncope-console.war!/BOOT-INF/lib/syncope-client-idrepo-common-ui-3.0.3-SNAPSHOT.jar!/org/apache/syncope/client/ui/commons/markup/html/form/AjaxSpinnerFieldPanel.html
<input wicket:id="spinner"/>, index = 0, current =  '<input wicket:id="spinner"/>' (line 0, column 0)]]]
```

PR #580, opened 2023-12-19
[SYNCOPE-1796] Implemented verification of access tokens issued by Microsoft Entra (formerly Azure)

philipptrenz:
Microsoft Entra uses self-issued access tokens, e.g. for SCIM provisioning requests, to authenticate on external systems. To authorise these requests in Syncope, they must be verified against public keys published by Microsoft in a JSON Web Key Set (JWKS).

This PR adds a `MSEntraAccessTokenJWSVerifier` implementation, which loads a JWKS provided by Microsoft from the corresponding endpoint, caches the public keys for the recommended duration and refetches them afterwards. The `MSEntraJWTSSOProvider` implementation allows to verify and authorize requests coming from Microsoft Entra based on the JSON Web Token (JWT) sent along in the request header.

### DIFF
--- a/core/provisioning-java/pom.xml
+++ b/core/provisioning-java/pom.xml
@@ -59,11 +59,6 @@ under the License.
     </dependency>
 
     <dependency>
-      <groupId>com.github.ben-manes.caffeine</groupId>
-      <artifactId>caffeine</artifactId>
-    </dependency>
-
-    <dependency>
       <groupId>org.apache.geronimo.javamail</groupId>
       <artifactId>geronimo-javamail_1.4_mail</artifactId>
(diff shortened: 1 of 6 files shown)


ilgrosso:
Thank you very much @philipptrenz 

I have found an ICLA matching your first name and surname, for username `philipp-trenz`: if you confirm that's you, then we are all set!

ilgrosso:
Even better, just found this older PR of yours, which includes ICLA ref: https://github.com/apache/syncope/pull/440

ilgrosso (approved):
The code looks very good and also features nice coverage

PR #581, opened 2023-12-19
[SYNCOPE-1797] Compatibility of SCIM 2.0 requests from Microsoft Entra

philipptrenz:
Following the [SCIM 2.0 specification](https://www.rfc-editor.org/rfc/rfc7643#section-4.1.1), the singular attribute `active` is specified of type Boolean. Deviating from this, some cloud providers (namely Microsoft Entra) on PATCH requests send this as a string value of `"True"` or `"False"`. 

To ensure compatibility, with this PR string representations (`"true"` and `"false"`, case insensitive) also get interpreted as Boolean values.

(Apache id: philipp-trenz)

### DIFF
--- a/ext/scimv2/logic/pom.xml
+++ b/ext/scimv2/logic/pom.xml
@@ -61,7 +61,19 @@ under the License.
       <groupId>org.antlr</groupId>
       <artifactId>antlr4-runtime</artifactId>
     </dependency>
-    
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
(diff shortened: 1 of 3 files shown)


ext/scimv2/logic/pom.xml
```diff
@@ -61,7 +61,19 @@ under the License.
       <groupId>org.antlr</groupId>
       <artifactId>antlr4-runtime</artifactId>
     </dependency>
-    
+
```
ilgrosso, line 64: Only `mockito-junit-jupiter`  should be needed.
Please remove `mockito-core`.
Also, remove the blank line after the `mockito-junit-jupiter` block

ilgrosso (changes requested):

ext/scimv2/logic/src/main/java/org/apache/syncope/core/logic/SCIMDataBinder.java
```diff
@@ -815,6 +815,17 @@ public Pair<UserUR, StatusR> toUserUpdate(final UserTO before, final SCIMPatchOp
 
             case "active":
                 if (!CollectionUtils.isEmpty(op.getValue())) {
+
+                    // Workaround for Microsoft Entra being not SCIM compliant on PATCH requests
+                    if (op.getValue().get(0) instanceof String) {
+                        String a = (String) op.getValue().get(0);
+                        if (a.equalsIgnoreCase("true")) {
```
ilgrosso, line 822: Can you change this whole block by 

```java
op.setValue(List.of(BooleanUtils.toBoolean(a)));
```

ilgrosso (approved):

PR #640, opened 2024-03-06
[SYNCOPE-1809] Remove uidOnCreate attribute on delete propagation

andrea-patricelli:


fit/core-reference/src/test/java/org/apache/syncope/fit/core/UserIssuesITCase.java
```diff
@@ -1634,4 +1640,72 @@ public void issueSYNCOPE1793() {
                 resource(RESOURCE_NAME_NOPROPAGATION).action(ResourceAssociationAction.ASSIGN).build());
         assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
     }
+
+    @Test
+    public void issue1809() throws IOException {
```
ilgrosso, line 1645: this should be named `issueSYNCOPE1809` (see others)

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/AbstractPropagationTaskExecutor.java
```diff
@@ -383,6 +383,20 @@ protected Uid delete(
 
             connector.delete(objectClass, uid, null, propagationAttempted);
             result = uid;
+            // SYNCOPE-1809 remove uidOnCreate attribute, if any
```
ilgrosso, line 386: are such comments really needed? We've got `git blame`

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/AbstractPropagationTaskExecutor.java
```diff
@@ -383,6 +383,20 @@ protected Uid delete(
...
+            taskInfo.getResource()
+                    .getProvisionByAnyType(taskInfo.getAnyType())
+                    .filter(provision -> provision.getUidOnCreate() != null)
+                    .ifPresent(provision -> {
+                        LOG.debug("Removing uidOnCreate [{}] attribute from [{}] on delete",
```
ilgrosso, line 391: either remove this logging statement or add it for the other case when `getUidOnCreate()` is used in this class

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/AbstractPropagationTaskExecutor.java
andrea-patricelli: No, going to remove.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/AbstractPropagationTaskExecutor.java
andrea-patricelli: Added logging also in the other section.

PR #769, opened 2024-07-01
[SYNCOPE-1822] added support for SCIM user extension

SamuelGaro:


ext/scimv2/scim-rest-api/src/main/java/org/apache/syncope/ext/scimv2/api/data/SCIMUser.java
```diff
@@ -80,6 +80,9 @@ public class SCIMUser extends SCIMResource {
     @JsonProperty("urn:ietf:params:scim:schemas:extension:enterprise:2.0:User")
     private SCIMEnterpriseInfo enterpriseInfo;
 
+    @JsonProperty("urn:ietf:params:scim:schemas:extension:syncope:2.0:User")
```
ilgrosso, line 83: what if one changes the urn?

ilgrosso (changes requested):

ilgrosso (approved):

PR #855, opened 2024-10-02
Switching crosschecks to JDK 24-ea

ilgrosso:


ilgrosso:
At the moment at least ByteBuddy and Mockito are not working with JDK > 23

PR #884, opened 2024-10-31
SYNCOPE-1835: Support Credential Criteria for LDAP authentication

mmoayyed:
See https://issues.apache.org/jira/browse/SYNCOPE-1835

image

PS Also fixes a small bug wrt to collecting the DN attribute.


ilgrosso (approved):
LGTM thank you @mmoayyed !

PR #901, opened 2024-11-08
Prepare 4.0.0-M0

ilgrosso:
The purpose of this PR is to attempt to remove any obstacle to release at least the first milestone for Syncope 4.0


ilgrosso:
It seems like [CXF 4.1.0 is approaching](https://lists.apache.org/thread/movf2qso9qhrnh9lwx0cfyg11ds11m32), so this PR does not seem to be needed any more.

PR #916, opened 2024-11-22
Remove unused exceptions from the throws clause

mmoayyed:


ilgrosso:
@mmoayyed you can run:

```
mvn checkstyle:check
```

before push, to check just for checkstyle

ilgrosso (approved):

PR #955, opened 2025-01-16
[SYNCOPE-1854] Propagate changes for user update while pending approval

leo-ccin2p3:
allow propagation to external resources after form validation

### DIFF
--- a/ext/flowable/flowable-bpmn/src/main/java/org/apache/syncope/core/flowable/impl/FlowableUserWorkflowAdapter.java
+++ b/ext/flowable/flowable-bpmn/src/main/java/org/apache/syncope/core/flowable/impl/FlowableUserWorkflowAdapter.java
@@ -324,6 +324,11 @@ public class FlowableUserWorkflowAdapter extends AbstractUserWorkflowAdapter imp
                 Optional.ofNullable(propByResBeforeUpdate).orElse(propByRes),
                 Optional.ofNullable(propByLinkedAccountBeforeUpdate).orElse(propByLinkedAccount));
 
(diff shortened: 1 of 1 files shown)


ilgrosso:
Hi @leo-ccin2p3 can you please illustrate the rationale behind this proposed change?

leo-ccin2p3:
Hi @ilgrosso 
In my BPMN workflow, I have implemented the ability to update certain attributes on a user which is already in status "updateApproved" (after the update the user will then stay in status "updateApprove"). 
This patch allows to propagate the change to the external resource in this case
Hope it's clear enough

ilgrosso:
Thanks for explanation @leo-ccin2p3 
Please go ahead and open an issue on https://issues.apache.org/jira/ with the description as above (or even with more details if possible).
Once we have a JIRA issue, we will update this PR's title to include a reference to that.

Thank you!

ilgrosso:
I saw the JIRA issue, thanks.

If this is your very first contribution to The Apache Software foundation, please send your [ICLA](https://www.apache.org/licenses/contributor-agreements.html#clas). thank you.

ilgrosso:
Adding a new test case or adjusting one of existing test cases under https://github.com/apache/syncope/blob/master/fit/core-reference/src/test/java/org/apache/syncope/fit/core/UserRequestITCase.java to check for this change would be nice.

ilgrosso:
Thank you very much @leo-ccin2p3 
I've also cherry-picked this to `3_0_X` as https://github.com/apache/syncope/commit/16efca5d2252669080e17280508c386e2609fe84

PR #1004, opened 2025-03-04
Chore: simplify stream API calls

mmoayyed:
Update relevant stream API calls and chains use `noneMatch`, `peek` and simplified versions of `forEach` and collectors.

### DIFF
--- a/client/am/console/src/main/java/org/apache/syncope/client/console/commons/AMRealmPolicyProvider.java
+++ b/client/am/console/src/main/java/org/apache/syncope/client/console/commons/AMRealmPolicyProvider.java
@@ -77,7 +77,7 @@ public class AMRealmPolicyProvider extends IdRepoRealmPolicyProvider {
                 new PropertyModel<>(realmTO, "authPolicy"),
                 false);
         authPolicy.setChoiceRenderer(new PolicyRenderer(authPolicies));
-        authPolicy.setChoices(authPolicies.keySet().stream().collect(Collectors.toList()));
+        authPolicy.setChoices(new ArrayList<>(authPolicies.keySet()));
         ((AbstractSingleSelectChoice<?>) authPolicy.getField()).setNullValid(true);
         view.add(authPolicy);
 
@@ -89,7 +89,7 @@ public class AMRealmPolicyProvider extends IdRepoRealmPolicyProvider {
                 new PropertyModel<>(realmTO, "ticketExpirationPolicy"),
                 false);
         ticketExpirationPolicy.setChoiceRenderer(new PolicyRenderer(ticketExpirationPolicies));
-        ticketExpirationPolicy.setChoices(ticketExpirationPolicies.keySet().stream().collect(Collectors.toList()));
+        ticketExpirationPolicy.setChoices(new ArrayList<>(ticketExpirationPolicies.keySet()));
         ((AbstractSingleSelectChoice<?>) ticketExpirationPolicy.getField()).setNullValid(true);
         view.add(ticketExpirationPolicy);
     }
--- a/client/idm/console/src/main/java/org/apache/syncope/client/console/status/AnyStatusDirectoryPanel.java
+++ b/client/idm/console/src/main/java/org/apache/syncope/client/console/status/AnyStatusDirectoryPanel.java
@@ -378,7 +378,7 @@ public class AnyStatusDirectoryPanel
                                     actual instanceof GroupTO);
                             statusBean.setLinked(false);
                             return statusBean;
-                        }).collect(Collectors.toList()));
+                        }).toList());
 
                 statusBeans.sort(comparator);
             }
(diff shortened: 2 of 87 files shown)


client/am/console/src/main/java/org/apache/syncope/client/console/commons/AMRealmPolicyProvider.java
```diff
@@ -77,7 +77,7 @@ public void add(final RealmTO realmTO, final RepeatingView view) {
                 new PropertyModel<>(realmTO, "authPolicy"),
                 false);
         authPolicy.setChoiceRenderer(new PolicyRenderer(authPolicies));
-        authPolicy.setChoices(authPolicies.keySet().stream().collect(Collectors.toList()));
+        authPolicy.setChoices(new ArrayList<>(authPolicies.keySet()));
```
ilgrosso, line 80: Can you provide some reference to explain that in general

```java
new ArrayList<>(collection);
```

shall be preferred over

```java
collection.stream().collect(Collectors.toList())
```

?

Thanks.

client/idm/console/src/main/java/org/apache/syncope/client/console/status/AnyStatusDirectoryPanel.java
```diff
@@ -378,7 +378,7 @@ protected List<StatusBean> getStatusBeans(final long first, final long count) {
                                     actual instanceof GroupTO);
                             statusBean.setLinked(false);
                             return statusBean;
-                        }).collect(Collectors.toList()));
+                        }).toList());
```
ilgrosso, line 381: Switching from

```java
collect(Collectors.toList())
```

to

```java
toList()
```

with Wicket should be carefully considered, as the given list object could be used as `Model` for a `Component` and required to be mutable.

client/am/console/src/main/java/org/apache/syncope/client/console/commons/AMRealmPolicyProvider.java
mmoayyed: Sure. The following comes to mind:

`new ArrayList<>(collection)` directly copies elements from the given collection using the `addAll()` method. This is an optimized operation. The stream variant of course has the overhead of creating a stream, iterating over elements, and creating a collector. The proposed syntax is also more compact with fewer calls, and you can also immediately see the list implementation used, unlike the toList() call where the implementation is somewhat unclear unless you look up docs or are intimately familiar with what is returned. 

Streaming would be useful if the intention is to do extra filtering or other intermediate ops, or if parallel processing is required, etc.

ilgrosso: Thank you, it makes sense.

client/idm/console/src/main/java/org/apache/syncope/client/console/status/ReconTaskPanel.java
```diff
@@ -143,7 +143,7 @@ protected Iterator<String> getChoices(final String input) {
                                     ? RealmsUtils.buildRootQuery()
                                     : RealmsUtils.buildKeywordQuery(input)).getResult())
                             : List.<RealmTO>of()).stream().
-                            map(RealmTO::getFullPath).collect(Collectors.toList()).iterator();
+                            map(RealmTO::getFullPath).toList().iterator();
```
ilgrosso, line 146: Why changing to `.toList().iterator()` and not directly `.iterator()`?

mmoayyed: Good point, thank you. I have revised all instances of this.

ilgrosso (approved):

client/idm/console/src/main/java/org/apache/syncope/client/console/status/AnyStatusDirectoryPanel.java
ilgrosso: @mmoayyed have you had the chance to test via browsers the changes in Console and Enduser? I am afraid the current test suite does not cover all scenarios.

mmoayyed: I ran some brief smoke tests, but will do some more thorough testing today.

ilgrosso: Thank you!

mmoayyed: I have run a set of tests against Syncope console and end-user, and so far I have not seen any problems. Creating realms, users, etc and various operations on users are work OK. I have also passed through various screens in WA and they too appear to work normally.

ilgrosso: Ok thanks.
The checks are needed especially with the operations that are involving the changed classes, so for example searching for Users, Groups or Any Objects, merging Linked Accounts, perform some Reconciliation etc.

Any way, let's go ahead and merge, just keep in mind that some immutable-related issues might arise and we know where they are from.

ilgrosso:
@mmoayyed please resolve the conflicts, thanks

ilgrosso (approved):
Let's take a note to investigate future errors related to immutable collections used for Wicket model objects.

PR #1010, opened 2025-03-05
Fix: fix log parameter placeholder mismatches

mmoayyed:


ilgrosso:
Nice spot!
The fix should be however to add the `schema` parameter to the `LOG` call

client/idrepo/console/src/main/java/org/apache/syncope/client/console/panels/AnyDirectoryPanel.java
```diff
@@ -136,7 +136,7 @@ protected AnyDirectoryPanel(final String id, final Builder<A, E> builder, final
                     try {
                         return schemaRestClient.<PlainSchemaTO>read(SchemaType.PLAIN, schema);
                     } catch (SyncopeClientException e) {
-                        LOG.warn("Could not read plain schema, ignoring", e);
+                        LOG.warn("Could not read plain schema {}, ignoring {}", schema, e);
```
ilgrosso, line 139: Shouldn't this be

```java
LOG.warn("Could not read plain schema {}, ignoring", schema, e);
```

?

PR #1087, opened 2025-05-26
[SYNCOPE-1886] move must to filter to improve query performance

andrea-patricelli:


ilgrosso:
Q 1: can you apply the same changes to `OpenSearchAnySearchDAO` ?
Q 2: is it correct that this PR is still draft?

andrea-patricelli:
> Q 1: can you apply the same changes to `OpenSearchAnySearchDAO` ?

Sure, let me work on it

> Q 2: is it correct that this PR is still draft?

Because I would try on some real-life environment first and I'm exploring other low-impact improvements (I think that we're done BTW).

> Q 3: is there any documentation available to support this change?

Yes, you can refer to [this](https://www.elastic.co/docs/explore-analyze/query-filter/languages/querydsl#query-filter-context) that explains query and filter context (used by filter) and [this blog post](https://toxigon.com/elasticsearch-query-optimization-tips) that, besides other useful tips, answers to the question "Should I use queries or filters for exact match conditions?"

ilgrosso:
@andrea-patricelli I've seen `SYNCOPE-1886`, thank you.
Would you mind adding the links you've reported in the comment above also in the issue description on JIRA, possibly with some description, to give more rationale about the proposed change?

PR #1125, opened 2025-07-03
[SYNCOPE-1889] improve group search using SCIM extension

SamuelGaro:


core/idrepo/logic/src/main/java/org/apache/syncope/core/logic/UserLogic.java
```diff
@@ -188,6 +189,17 @@ public Pair<Integer, List<UserTO>> search(
         return Pair.of(count, result);
     }
 
+    @PreAuthorize("hasRole('" + IdRepoEntitlement.USER_SEARCH + "')")
```
ilgrosso: I am not exactly a fan of adding this method in `UserLogic`.

What happens if you use directly `GroupDAO#findUMemberships` in `SCIMDataBinder`?

SamuelGaro: I think is fine, I proceed with the modification

ilgrosso:
@SamuelGaro as you can see, there is a test failure:

```
SCIMITCase.search:388 expected: <true> but was: <false>
```

PR #1141, opened 2025-07-24
[SYNCOPE-1898] Reset password with WA if mustChangePassword is true

alberto-bogi:


wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/WAPropertySourceLocator.java
```diff
@@ -124,6 +127,12 @@ public PropertySource<?> locate(final Environment environment) {
             properties.putAll(index(map, prefixes));
         });
 
+        LOG.debug("Configure properties for CAS password manager on syncope:");
```
ilgrosso: This would be the sole LOG of such a type in this class, please remove

wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/mapping/PropertySourceMapper.java
```diff
@@ -37,7 +37,7 @@ public abstract class PropertySourceMapper {
 
     protected static final Logger LOG = LoggerFactory.getLogger(PropertySourceMapper.class);
 
-    protected static Map<String, Object> prefix(final String prefix, final Map<String, Object> map) {
+    public static Map<String, Object> prefix(final String prefix, final Map<String, Object> map) {
```
ilgrosso: Why it is needed to change this visibility?

fit/wa-reference/src/test/java/org/apache/syncope/fit/AbstractITCase.java
```diff
@@ -231,6 +231,19 @@ protected static String extractWAExecution(final String body) {
         return execution.get();
     }
 
+    protected static String extractWACSRF(final String body) {
```
ilgrosso: Why placing this method in this class? If used only by `OIDCC4UIITCase` it should stay there

ilgrosso (changes requested):

wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/mapping/PropertySourceMapper.java
alberto-bogi: Because `prefix` method is defined in package `org.apache.syncope.wa.bootstrap.mapping`, but i use this method also in `WAPropertySourceLocator` that is defined in package `org.apache.syncope.wa.bootstrap`. I thought it might be correct to use the same method used for the other mappings

fit/wa-reference/src/test/java/org/apache/syncope/fit/AbstractITCase.java
alberto-bogi: I use this method also for `SAML2SP4UIITCase`. It's a form input required in CAS password management flow

ilgrosso: Ok, so in `AbstractUIITCase` not in `AbstractITCase` then

wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/WAPropertySourceLocator.java
```diff
@@ -124,6 +127,12 @@ public PropertySource<?> locate(final Environment environment) {
...
+        SyncopePasswordManagementProperties pm = new SyncopePasswordManagementProperties();
```
ilgrosso: This seems overkill, just use plain

```java
properties.put(
"cas.authn.pm.syncope.url", 
StringUtils.substringBefore(syncopeClient.getAddress(), "/rest"));
```

and you will also avoid changing the visibility of `PropertySourceMapper#prefix`

wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/mapping/PropertySourceMapper.java
ilgrosso: See my other comment for the change in `WAPropertySourceLocator`: no need to make it such complex

wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/WAPropertySourceLocator.java
alberto-bogi: Ok perfect. Thank you!

wa/bootstrap/src/main/java/org/apache/syncope/wa/bootstrap/WAPropertySourceLocator.java
```diff
@@ -124,6 +124,9 @@ public PropertySource<?> locate(final Environment environment) {
             properties.putAll(index(map, prefixes));
         });
 
+        properties.put("cas.authn.pm.syncope.url",
+                StringUtils.substringBefore(syncopeClient.getAddress(), "/rest"));
+
         Set<String> customClaims = syncopeClient.getService(WAClientAppService.class).list().stream().
```
ilgrosso, line 130: Remove this LOG statement

ilgrosso (changes requested):
I don't see any more the method `extractWACSRF` defined, I was expecting it in `AbstractUIITCase` no?

ilgrosso:
@alberto-bogi I am also seeing

```
SAML2SP4UIITCase>AbstractUIITCase.passwordManagementEnduser:176 » SyncopeClient EntityExists [A violation of the constraint imposed by a unique index or a unique constraint occurred]
```

alberto-bogi:
> @alberto-bogi I am also seeing
> 
> ```
> SAML2SP4UIITCase>AbstractUIITCase.passwordManagementEnduser:176 » SyncopeClient EntityExists [A violation of the constraint imposed by a unique index or a unique constraint occurred]
> ```

I''m seeing that. I create a user with username `mustChangePassword` before executing `passwordManagement` method, and i delete this after. This in method `passwordManagementConsole` and `passwordManagementEnduser`. What am I doing wrong?

ilgrosso:
@alberto-bogi 

```
You have 1 Checkstyle violation
```

ilgrosso (approved):

alberto-bogi:
> Thank you very much @alberto-bogi !

Thank you for your help!

PR #1142, opened 2025-07-24
[SYNCOPE-1898] Reset password with WA if mustChangePassword is true

ilgrosso:
Creating this as draft because we need to understand how to make the new test cases work against CAS 7.3.x


ilgrosso:
Added tmp properties in https://github.com/apache/syncope/pull/1142/commits/dfb82a1321908ad80eaf4c8820947c2fcb7cd76f to let the tests work until [SYNCOPE-1901](https://issues.apache.org/jira/browse/SYNCOPE-1901) gets implemented

PR #1159, opened 2025-08-21
Separate membership attribute on __ACCOUNT__ and ldapGroups attribute for connector

markusokon:
Currently LDAPMembershipPropagationActions uses the same attribute for writing the result into the propagation data and fetching the preexisting group memberships of the user object in LDAP. This leads to the beforeObj.getAttributeByName() call never returning any groups in the default case and therefore Syncope trying to edit groups it doesn't own/control.

This is fixed in this pull request by separating the attribute name used into one which the connector receives, containing all the group memberships after the execution, and the attribute name which is used to get all current memberships from the LDAP object.

Furthermore I added a performance optimization by searching for the groups managed by Syncope only once and not potentially hundreds of times.

### DIFF
--- a/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/LDAPMembershipPropagationActions.java
+++ b/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/LDAPMembershipPropagationActions.java
@@ -133,7 +133,7 @@ public class LDAPMembershipPropagationActions implements PropagationActions {
                     orElseThrow(() -> new NotFoundException("User " + taskInfo.getEntityKey()));
             Set<String> groups = new HashSet<>();
 
-            // for each user group assigned to the resource of this task, compute and add the group's 
+            // for each user group assigned to the resource of this task, compute and add the group's
             // connector object link
             userDAO.findAllGroupKeys(user).stream().
                     map(groupDAO::findById).flatMap(Optional::stream).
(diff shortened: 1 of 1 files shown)


markusokon:
Hi @ilgrosso 

it's currently under discussion if this code also requires a CCLA to be signed and I will give you an update ASAP.

ilgrosso:
Hi @markusokon it's been about a month since last update: please let us know if there is any news, thanks.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/LDAPMembershipPropagationActions.java
```diff
@@ -76,11 +76,25 @@ public class LDAPMembershipPropagationActions implements PropagationActions {
...
      *
-     * @return the name of the attribute used to keep track of group memberships
+     * @return the name of the attribute used to keep track of group memberships on the __ACCOUNT__ object
      */
     protected String getGroupMembershipAttrName() {
+        return "memberOf";
+    }
+
+    @Override
+    public Set<String> moreAttrsToGet(Optional<PropagationTaskInfo> taskInfo, Provision provision) {
```
ilgrosso: `final` keyword missing for method's arguments

ilgrosso (changes requested):
I'd suggest to perform at least a full build by simple

```
mvn -T 1C clean install
```

from the root folder.

ilgrosso:
@markusokon do you have any news to fix the failing integration test?

PR #1176, opened 2025-09-12
[SYNCOPE-1912] Check LinkedAccounts for password propagation after User update

valerio-crescia:


core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/DefaultPropagationManager.java
```diff
@@ -236,11 +238,16 @@ public List<PropagationTaskInfo> getUserUpdateTasks(
     public List<PropagationTaskInfo> getUserUpdateTasks(final UserWorkflowResult<Pair<UserUR, Boolean>> wfResult) {
         UserUR userUR = wfResult.getResult().getLeft();
 
+        List<String> linkedAccountsResources = userUR.getLinkedAccounts().stream()
+            .map(LinkedAccountUR::getLinkedAccountTO)
+            .filter(linkedAccountTO -> linkedAccountTO.getPassword() != null)
```
ilgrosso: this should be

```java
linkedAccountTO -> linkedAccountTO != null 
  && linkedAccountTO.getPassword() != null
```

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/propagation/DefaultPropagationManager.java
```diff
@@ -236,11 +238,16 @@ public List<PropagationTaskInfo> getUserUpdateTasks(
...
+            .map(LinkedAccountTO::getResource).collect(Collectors.toList());
```
ilgrosso: need to add `.distinct()` to the stream to avoid to repeat the same Resource several times

PR #1198, opened 2025-09-29
[SYNCOPE-1918] Avoided dynamic membership deletion after push task execution

SamuelGaro:


core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/pushpull/AbstractPushResultHandler.java
```diff
@@ -179,6 +181,10 @@ protected void link(final Any<?> any, final boolean unlink, final ProvisioningRe
                 operation(unlink ? PatchOperation.DELETE : PatchOperation.ADD_REPLACE).
                 value(profile.getTask().getResource().getKey()).build());
 
+        if (any instanceof Group && ((Group) any).getUDynMembership() != null) {
```
ilgrosso: You should do the same for `adynMembershipConds`

Moreover, it could make sense to centralize the changes you are introducing in a single method like as

```java
protected void copyDynMembershipConds(final Any<?> any, final AnyUR req) {
if (any instanceof Group && ((Group) any).getUDynMembership() != null) {
            ((GroupUR) req).setUDynMembershipCond(((Group) any).getUDynMembership().getFIQLCond());

            // also copy aDynMembershipConds
        }
}
```

ilgrosso (changes requested):

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/pushpull/AbstractPushResultHandler.java
```diff
@@ -201,11 +207,24 @@ protected void assign(final Any<?> any, final Boolean enabled, final Provisionin
...
 
+        copyDynMembershipConds(any, req);
+
         update(req);
 
         provision(any, enabled, result);
     }
 
+    protected void copyDynMembershipConds(final Any<?> any, final AnyUR req) {
+        if (any instanceof Group
```
ilgrosso: What if `.getUDynMembership()` is null but `.getUDynMembership()` is not empty?

ilgrosso (changes requested):

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/pushpull/AbstractPushResultHandler.java
SamuelGaro: I see, I'm going to fix

ilgrosso: I think you will need a structured `if`:

```java
 if (any instanceof Group) {
   Optional.ofNullable((Group) any).getUDynMembership()).
ifPresent(udc ->  ((GroupUR) req).setUDynMembershipCond(udc.getFIQLCond()));

  ((Group) any).getADynMemberships().forEach(aDynGroupMembership ->
                    ((GroupUR) req).getADynMembershipConds().put(
                            aDynGroupMembership.getAnyType().getKey(), aDynGroupMembership.getFIQLCond()));
 }
```

ilgrosso (approved):

PR #1350, opened 2026-04-15
[SYNCOPE-1962] Replace all columns type unknown(2014) with timestamp, edit table names and indexes with length more than 30 characters and change column jwks in table OIDCOpEntity from clob to blob

valerio-crescia:


core/persistence-jpa-upgrader/src/main/resources/schema-oracle.xml
```diff
@@ -751,9 +751,9 @@ under the License.
             <column name="subjectType" type="varchar" size="20"/>
             <column name="supportedGrantTypes" type="clob" size="-1"/>
             <column name="supportedResponseTypes" type="clob" size="-1"/>
-            <column name="tokenEndpointAuthenticationMethod" type="varchar" size="20"/>
+            <column name="tokenEndpointAuthenticationMet" type="varchar" size="20"/>
```
ilgrosso, line 754: Change column names shall not be included, as it might break the JPA mapping

core/persistence-jpa-upgrader/src/main/resources/schema-oracle.xml
```diff
@@ -751,9 +751,9 @@ under the License.
...
             <column name="userInfoEncryptedResponseAlg" type="varchar" size="20"/>
-            <column name="userInfoEncryptedResponseEncoding" type="varchar" size="20"/>
+            <column name="userInfoEncryptedResponseEncod" type="varchar" size="20"/>
```
ilgrosso, line 756: see above

ilgrosso (changes requested):

core/persistence-jpa-upgrader/src/main/resources/schema-oracle.xml
valerio-crescia: The tool stops with the tables name length more than 30 characters. I don't think it breaks something because the actual tablenames on Syncope 4.0 are like that on Oracle

ilgrosso: ok

PR #1352, opened 2026-04-19
SYNCOPE-1744: restore notification template context after user delete

ozimakov:
### Problem
For notifications bound to `[LOGIC]:[UserLogic]:[]:[delete]:[SUCCESS]`, templates that reference the subject user (e.g. `${user.getPlainAttr("u_email").get().values[0]}`) were evaluated with an empty user: after delete, `userDAO.findById` no longer returns the entity, so the JEXL context never received a user variable. The before snapshot from `LogicInvocationHandler` already held a full `UserTO`, but it was not used when the persisted entity was gone.

### Solution
In `DefaultNotificationManager.createTasks`, when no Any can be loaded from the persistence layer but before is a `UserTO`, `GroupTO`, or `AnyObjectTO`, populate the corresponding JEXL variables (user, group, anyObject) from that snapshot so mail templates keep working for delete success events.

### Changes
`DefaultNotificationManager.java`: fallback to before transfer objects when any is empty.
`DefaultNotificationManagerTest.java`: unit test with `userDAO.findById` empty and a template using plain attributes (same JEXL style as existing mail template tests).

### Testing
```
mvn -pl core/provisioning-java -Dtest=DefaultNotificationManagerTest test
mvn -pl core/provisioning-java checkstyle:check
mvn clean test -pl core/provisioning-java -am
mvn test (full reactor)
```

### References
https://issues.apache.org/jira/browse/SYNCOPE-1744

### DIFF
--- a/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/notification/DefaultNotificationManager.java
+++ b/core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/notification/DefaultNotificationManager.java
@@ -343,27 +343,33 @@ public class DefaultNotificationManager implements NotificationManager {
                     jexlVars.put("output", output);
                     jexlVars.put("input", input);
 
-                    any.ifPresent(a -> {
-                        switch (a) {
-                            case User user ->
-                                jexlVars.put("user", userDataBinder.getUserTO(user, true));
-                            case Group group ->
-                                jexlVars.put("group", groupDataBinder.getGroupTO(group, true));
-                            case AnyObject anyObject ->
-                                jexlVars.put("anyObject", anyObjectDataBinder.getAnyObjectTO(anyObject, true));
-                            default -> {
-                            }
-                        }
-                    });
-                    if (any.isEmpty()) {
-                        if (before instanceof UserTO userTO) {
-                            jexlVars.put("user", userTO);
-                        } else if (before instanceof GroupTO groupTO) {
-                            jexlVars.put("group", groupTO);
-                        } else if (before instanceof AnyObjectTO anyObjectTO) {
-                            jexlVars.put("anyObject", anyObjectTO);
-                        }
-                    }
+                    any.ifPresentOrElse(
+                            a -> {
+                                switch (a) {
+                                    case User user ->
+                                        jexlVars.put("user", userDataBinder.getUserTO(user, true));
+                                    case Group group ->
+                                        jexlVars.put("group", groupDataBinder.getGroupTO(group, true));
+                                    case AnyObject anyObject ->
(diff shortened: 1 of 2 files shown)


ilgrosso:
Hi @ozimakov thanks for this PR.
If you haven't done this before, please go ahead and submit [your ICLA](https://www.apache.org/licenses/contributor-agreements.html#clas) to the ASF, thank you.

ozimakov:
> Hi @ozimakov thanks for this PR. If you haven't done this before, please go ahead and submit [your ICLA](https://www.apache.org/licenses/contributor-agreements.html#clas) to the ASF, thank you.

Thanks for the heads-up and for your continued work. I've submitted the documentation.

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/notification/DefaultNotificationManager.java
```diff
@@ -355,6 +355,15 @@ public List<NotificationTask> createTasks(
                             }
                         }
                     });
+                    if (any.isEmpty()) {
```
ilgrosso: Why not moving these statements above, e.g. change lines 346 and below from

```java
any.ifPresent(a -> {
...
});
```

to

```java
                    any.ifPresentOrElse(
                            a -> {
                                switch (a) {
                                    case User user ->
                                        jexlVars.put("user", userDataBinder.getUserTO(user, true));
                                    case Group group ->
                                        jexlVars.put("group", groupDataBinder.getGroupTO(group, true));
                                    case AnyObject anyObject ->
                                        jexlVars.put("anyObject", anyObjectDataBinder.getAnyObjectTO(anyObject, true));
                                    default -> {
                                    }
                                }
                            },
                            () -> {
                                switch (before) {
                                    case UserTO userTO ->
                                        jexlVars.put("user", userTO);
                                    case GroupTO groupTO ->
                                        jexlVars.put("group", groupTO);
                                    case AnyObjectTO anyObjectTO ->
                                        jexlVars.put("anyObject", anyObjectTO);
                                    default -> {
                                    }
                                }
                            });
```

ilgrosso (changes requested):

core/provisioning-java/src/main/java/org/apache/syncope/core/provisioning/java/notification/DefaultNotificationManager.java
ozimakov: done with additional null check, just in case

ilgrosso (approved):

ilgrosso:
Thank you @ozimakov !

ilgrosso:
FYI I have cherry-picked your commit to fix the issue on branches `4_1_X` and `4_0_X` as well.

PR #1362, opened 2026-04-25
[SYNCOPE-1921] LDAPMembershipPropagationActions preserves external group memberships

ozimakov:
## Summary

- Override `moreAttrsToGet()` in `LDAPMembershipPropagationActions` to request `ldapGroups` when fetching the remote object before propagation
- Without this, `beforeObj` does not contain `ldapGroups` when it is not in the user mapping, causing the external-group preservation block (lines 154-169) to be a no-op

## Test plan

- [x] New IT `issueSYNCOPE1921` — same scenario as `issueSYNCOPE1473` but without `ldapGroups` in user mapping; fails without the fix, passes with it
- [x] Existing IT `issueSYNCOPE1473` passes (no regression)


ilgrosso (approved):

ilgrosso:
Thank you @ozimakov !

PR #1389, opened 2026-05-17
SYNCOPE-1758 Avoid CAS servlet reinitialization in WA

frhack:
This overrides CAS' configuration refresh listener in the WA starter so refresh still eagerly resolves beans, but no longer manually calls `DispatcherServlet.init()` without a `ServletConfig`.

The root cause is tracked upstream in CAS:
https://github.com/apereo/cas/pull/8974

Syncope 3.0.x currently depends on CAS 6.6.x, and that CAS release line is EOL, so this keeps Syncope WA protected without waiting for a usable upstream CAS release on the 6.6 line.

Changes:
- Add a WA-specific `casConfigurationEventListener` bean.
- Preserve CAS refresh/rebind behavior.
- Remove the special-case manual `DispatcherServlet` initialization.
- Add a regression test verifying that `dispatcherServlet` is resolved but not reinitialized.

Tests:
- `/tmp/maven/apache-maven-3.9.11/bin/mvn -pl wa/starter -am -Dtest=WACasConfigurationEventListenerTest -Dsurefire.failIfNoSpecifiedTests=false test`

PR #1390, opened 2026-05-17
SYNCOPE-1758 Avoid CAS servlet reinitialization in WA

frhack:
This is the `4_1_X` version of #1389, following the review feedback there.

The PR defines the WA-specific `casConfigurationEventListener` bean directly in `WAContext`, so it is picked up by Spring runtime, and keeps the same eager/non-lazy refresh listener contract as CAS.

The listener preserves CAS refresh/rebind behavior, but avoids manually invoking `DispatcherServlet.init()` during refresh. In Syncope WA that servlet has no `ServletConfig` at that point, which can trigger the `IllegalStateException` reported in SYNCOPE-1758.

Upstream CAS root cause fix: https://github.com/apereo/cas/pull/8974

Tested with:

```
/tmp/maven/apache-maven-3.9.11/bin/mvn -pl wa/starter -am -Dtest=WACasConfigurationEventListenerTest -Dsurefire.failIfNoSpecifiedTests=false test
```


ilgrosso:
https://github.com/apache/syncope/pull/1389 was closed, closing this one too.

Please update here in case the proposed change gets merged into CAS first

PR #1411, opened 2026-06-04
Tighten auth failure lockout threshold

massx1:
Before this change, the account policy threshold was enforced on attempt N + 1: with maxAuthenticationAttempts = 3, the user was still active after 3 failures and suspended only on the 4th failure. 

The old test encoded this by asserting failedLogins == 3 before executing one more failed login.

### DIFF
--- a/fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
+++ b/fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
@@ -20,6 +20,7 @@ package org.apache.syncope.fit.core;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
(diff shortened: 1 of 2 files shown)


fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
```diff
@@ -353,9 +348,10 @@ public void checkUserSuspension() {
             assertNotNull(e);
         }
 
-        assertEquals(3, USER_SERVICE.read(userKey).getFailedLogins());
+        userTO = USER_SERVICE.read(userTO.getKey());
+        assertEquals(2, userTO.getFailedLogins().intValue());
+        assertNotEquals("suspended", userTO.getStatus());
```
ilgrosso, line 353: The `status` value comes from the workflow definition in use, so it might vary.

The cleanest check here is

```
assertFalse(userTO.isSuspended();
```

ilgrosso (changes requested):

ilgrosso:
> Before this change, the account policy threshold was enforced on attempt N + 1: with maxAuthenticationAttempts = 3, the user was still active after 3 failures and suspended only on the 4th failure.

maxAuthenticationAttempts = 3 

means that the number of authentication attempts allowed to fail is 3, at maximum.

So the fact that  "the user was still active after 3 failures and suspended only on the 4th failure" is on purpose, not an error.

ilgrosso:
@massx1 please reach out to dev@ to discuss, when you are proposing such changes, thanks.

fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
massx1: I guess I have to do the same thing few rows below adding assertTrue(userTO.isSuspended());
```
userTO = USER_SERVICE.read(userTO.getKey());
assertNotNull(userTO);
assertNotNull(userTO.getFailedLogins());
assertEquals(3, userTO.getFailedLogins().intValue());
assertEquals("suspended", userTO.getStatus());
```

PR #1414, opened 2026-06-08
[SYNCOPE-1973] Avoid exposing authentication error details that enable user enumeration

massx1:
This change makes authentication failure details configurable to reduce user enumeration through error messages.
 
When detailed authentication errors are disabled, failed username/password authentication attempts return a generic error message in RESTHeaders.ERROR_INFO instead of exposing details such as an unknown username, invalid password, or suspended user status.

The default behavior is configured through security.authenticationError.exposeDetails and security.authenticationError.genericMessage.

### DIFF
--- a/fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
+++ b/fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
@@ -71,6 +71,7 @@ import org.apache.syncope.common.rest.api.beans.UserRequestQuery;
 import org.apache.syncope.common.rest.api.service.AnyObjectService;
 import org.apache.syncope.common.rest.api.service.SchemaService;
 import org.apache.syncope.common.rest.api.service.UserService;
+import org.apache.syncope.core.spring.security.SecurityProperties;
 import org.apache.syncope.fit.AbstractITCase;
 import org.junit.jupiter.api.Test;
 import org.springframework.jdbc.core.JdbcTemplate;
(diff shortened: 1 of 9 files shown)


ilgrosso:
@massx1 again, this needs a JIRA

fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
```diff
@@ -322,6 +323,31 @@ public void checkFailedLogins() {
         assertEquals(0, userService4.read(userKey).getFailedLogins());
     }
 
+    @Test
+    public void authenticationFailureErrorInfoIsGeneric() {
+        try {
```
ilgrosso: use `assertThrows`

fit/core-reference/src/test/java/org/apache/syncope/fit/core/AuthenticationITCase.java
```diff
@@ -322,6 +323,31 @@ public void checkFailedLogins() {
...
+                    e.getResponse().getHeaderString(RESTHeaders.ERROR_INFO));
+        }
+
+        UserCR userCR = UserITCase.getUniqueSample("genericAuthError@syncope.apache.org");
+        userCR.getRoles().add("User manager");
+
+        UserTO userTO = createUser(userCR).getEntity();
+        assertNotNull(userTO);
+
+        try {
```
ilgrosso: use `assertThrows`

ilgrosso (changes requested):

ilgrosso (approved):

PR #1419, opened 2026-06-09
[SYNCOPE-1975] Throttle password reset requests

massx1:
This change adds configurable throttling for anonymous password reset requests.

The throttling is applied before user lookup, so repeated requests for both existing and non-existing usernames are  handled consistently. 

When the configured threshold is exceeded, Syncope returns HTTP 429 Too Many Requests with a Retry-After header.

The feature is controlled by dedicated security properties.

The client address is included in the throttling key to avoid allowing an attacker who knows a username or email address to globally block password reset requests for that account.


ilgrosso:
@massx1 the PR looks ok now, please add some docs and we can merge it

ilgrosso (approved):

PR #1420, opened 2026-06-09
[SYNCOPE-1976] Add configurable REST rate limiting

massx1:
Adds optional REST rate limiting for Syncope Core APIs.

The feature allows administrators to configure request thresholds, time windows, lock periods, trusted proxies, and excluded client addresses (for example Console instances). 

When enabled, clients exceeding the configured threshold receive 429 Too Many Requests with a Retry-After header.

The feature is disabled by default.

Notes
This is intended as a lightweight L7 mitigation for excessive request rates against CXF REST endpoints, complementing existing endpoint-specific throttling such as login and password reset protection.


core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/RESTProperties.java
```diff
@@ -28,7 +31,79 @@ public class RESTProperties {
     @NestedConfigurationProperty
     private final ExecutorProperties batchExecutor = new ExecutorProperties();
 
+    @NestedConfigurationProperty
+    private final RateLimit rateLimit = new RateLimit();
```
ilgrosso: Please rename this class as `RateLimitProperties`

core/idrepo/rest-cxf/src/main/java/org/apache/syncope/core/rest/cxf/CXFRateLimitFilter.java
```diff
@@ -0,0 +1,177 @@
...
+
+    protected record RateLimitDecision(boolean allowed, long retryAfterSeconds) {
+    }
+
+    public record ClientWindow(long windowStartMillis, int count, long lockedUntilMillis) implements Serializable {
+
+        private static final long serialVersionUID = -473897805205955157L;
+    }
+
+    public static MutableConfiguration<String, ClientWindow> cacheConfiguration(final RESTProperties.RateLimit props) {
```
ilgrosso: Please move the cache definition in the Spring context.

I can imagine you did put it here to reuse it in test code: no issue with duplicated code for tests.

ilgrosso: Moreover, we have already `IdRepoRESTCXFTestContext` there.

ilgrosso (changes requested):
The filter is now defined, tested and added to the Spring context, but not actually configured in CXF.

You will need to add it in `IdRepoRESTCXFContext#restContainer` to do so.

ilgrosso:
@massx1 the PR looks ok now, can you add some notes to the reference guide about this feature?

ilgrosso (approved):

PR #1432, opened 2026-06-22
[SYNCOPE-1979] check also for not permitted words and schemas as substrings in default password rule

andrea-patricelli:


core/spring/src/main/java/org/apache/syncope/core/spring/policy/DefaultPasswordRule.java
```diff
@@ -89,8 +89,8 @@ public void setConf(final PasswordRuleConf conf) {
     protected void enforce(final String username, final String clearPassword, final Collection<String> notPermitted) {
         List<Rule> rules = PasswordGenerator.conf2Rules(conf);
         if (!notPermitted.isEmpty()) {
-            rules.add(new DictionaryRule(new WordListDictionary(new ArrayWordList(
-                    notPermitted.stream().distinct().sorted(Comparator.naturalOrder()).toArray(String[]::new), true)),
+            rules.add(new DictionarySubstringRule(new WordListDictionary(new ArrayWordList(
+                    notPermitted.stream().distinct().sorted(Comparator.naturalOrder()).toArray(String[]::new), false)),
```
ilgrosso: Why `false` by default here?

Since you are introducing case insensitive checks, I would expect that the policy rule has a flag to configure this, and also that such new flag is applied everywhere it makes sense.

core/spring/src/main/java/org/apache/syncope/core/spring/security/PasswordGenerator.java
```diff
@@ -96,8 +96,8 @@ public String getCharacters() {
 
         if (!conf.getWordsNotPermitted().isEmpty()) {
             conf.getWordsNotPermitted().sort(Comparator.naturalOrder());
-            rules.add(new DictionaryRule(new WordListDictionary(
-                    new ArrayWordList(conf.getWordsNotPermitted().toArray(String[]::new), true)), true));
+            rules.add(new DictionarySubstringRule((new WordListDictionary(
+                    new ArrayWordList(conf.getWordsNotPermitted().toArray(String[]::new), false))), true));
```
ilgrosso: See above.

fit/core-reference/src/test/java/org/apache/syncope/fit/core/PolicyITCase.java
```diff
@@ -482,4 +484,40 @@ public void issueSYNCOPE682() {
...
+        DefaultPasswordRuleConf defaultPasswordRuleConf =
+                POJOHelper.deserialize(originalRule.getBody(), DefaultPasswordRuleConf.class);
+        defaultPasswordRuleConf.getSchemasNotPermitted().add("firstname");
+        originalRule.setBody(POJOHelper.serialize(defaultPasswordRuleConf));
+        IMPLEMENTATION_SERVICE.update(originalRule);
+        try {
+            UserCR userCR = UserITCase.getUniqueSample("syncope1979@syncope.apache.org");
+            // 1. set password with not permitted word inside
+            userCR.setPassword("Notpermitted12345!");
+            try {
```
ilgrosso: Use `assertThrows()`, not the  `try / catch` block

fit/core-reference/src/test/java/org/apache/syncope/fit/core/PolicyITCase.java
```diff
@@ -482,4 +484,40 @@ public void issueSYNCOPE682() {
...
+                createUser(userCR);
+                fail("This should not happen");
+            } catch (SyncopeClientException e) {
+                assertEquals(ClientExceptionType.InvalidUser, e.getType());
+                assertTrue(e.getElements().iterator().next().startsWith("InvalidPassword"));
+            }
+            // 2. set password with not permitted schema inside
+            userCR.setPassword(userCR.getPlainAttr("firstname").get().getValues().getFirst() + "12345!");
+            try {
```
ilgrosso: same here

ilgrosso (changes requested):

core/spring/src/main/java/org/apache/syncope/core/spring/security/PasswordGenerator.java
andrea-patricelli: Before [this commit](https://github.com/apache/syncope/commit/343714fe7ed5dcc1d8303efa3733c9c09aae9d1a) the check was on substrings without additional flag, going to add it.

core/spring/src/main/java/org/apache/syncope/core/spring/policy/DefaultPasswordRule.java
andrea-patricelli: I assumed that case sensitivity was not so useful on password check, going to add the flag.

ilgrosso (approved):

core/spring/src/main/java/org/apache/syncope/core/spring/security/PasswordGenerator.java
```diff
@@ -95,9 +96,13 @@ public String getCharacters() {
         }
 
         if (!conf.getWordsNotPermitted().isEmpty()) {
-            conf.getWordsNotPermitted().sort(Comparator.naturalOrder());
-            rules.add(new DictionaryRule(new WordListDictionary(
-                    new ArrayWordList(conf.getWordsNotPermitted().toArray(String[]::new), true)), true));
+            WordListDictionary wld = new WordListDictionary(
```
ilgrosso, line 99: Case insensitive ordering shall be applied here too

ilgrosso (changes requested):

ilgrosso (approved):

PR #1438, opened 2026-06-25
[SYNCOPE-1981] Search audit events by who

ozimakov:
[SYNCOPE-1981](https://issues.apache.org/jira/browse/SYNCOPE-1981)

> Split out from [SYNCOPE-1978](https://issues.apache.org/jira/browse/SYNCOPE-1978) per review: this PR adds searching by **who performed** the action. Matching audit events by the **username of the affected entity** (the actual subject of SYNCOPE-1978) will be addressed separately.

### What
Adds the ability to search audit events by the principal that performed them. A new `who` query parameter on `AuditQuery` matches the `AuditEvent.who` column by **exact value** and may be **repeated to match any of the given values** (OR):

- single — `?who=jsmith`
- multiple, OR-ed — `?who=jsmith&who=admin`

It composes (AND) with the existing audit search filters (`entityKey`, `type`, `category`, `op`, `outcome`, `before`/`after`).

### Why
There is no way to answer "what did administrator _X_ do". This is useful to retrieve all actions performed by a given principal — in particular when that account has since been deleted and its entity key is no longer available. Syncope 3.0 allowed an approximation by abusing `entityKey=<username>`; the 4.0 audit refactor correctly made `entityKey` match only the affected entity's UUID, removing that side effect.

### Implementation
The filter is threaded through `AuditServiceImpl` → `AuditLogic` → the `AuditEventDAO` interface and all of its implementations: JPA, Neo4j, Elasticsearch and OpenSearch.

- JPA: `who IN (?, ...)` with bound parameters;
- Neo4j: `n.who IN $who` with a bound list parameter;
- Elasticsearch / OpenSearch: a `term` query per value, OR-ed via `bool`/`should`.

All values are bound as query parameters (SQL/Cypher) or passed as structured term queries (Elasticsearch/OpenSearch), so there is no injection surface.

### Tests
New integration tests in `AuditITCase` cover exact match, multiple values (OR), no-match, and that the match is exact (a `*` is treated literally, not as a wildcard), composed with the other filters. Manually verified end-to-end against embedded PostgreSQL and Neo4j.

### DIFF
--- a/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAuditEventDAO.java
+++ b/core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAuditEventDAO.java
@@ -24,6 +24,7 @@ import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.syncope.common.lib.to.AuditEventTO;
 import org.apache.syncope.common.lib.types.OpEvent;
@@ -32,6 +33,7 @@ import org.apache.syncope.core.persistence.api.entity.AuditEvent;
 import org.apache.syncope.core.persistence.jpa.entity.JPAAuditEvent;
 import org.springframework.data.domain.Pageable;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.util.CollectionUtils;
 
 public class JPAAuditEventDAO implements AuditEventDAO {
 
@@ -55,6 +57,17 @@ public class JPAAuditEventDAO implements AuditEventDAO {
                         append("inputs LIKE '%\"key\":\"").append(entityKey).append("\"%' OR ").
                         append("output LIKE '%\"key\":\"").append(entityKey).append("\"%' OR ").
                         append("throwable LIKE '%\"key\":\"").append(entityKey).append("\"%')");
+            }
+            return this;
+        }
+
+        public AuditEventCriteriaBuilder who(final Set<String> who, final List<Object> parameters) {
+            if (!CollectionUtils.isEmpty(who)) {
+                query.append(andIfNeeded()).append("who IN (").
+                        append(who.stream().
+                                map(value -> "?" + setParameter(parameters, value)).
+                                collect(Collectors.joining(", "))).
+                        append(")");
             }
             return this;
         }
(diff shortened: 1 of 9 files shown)


core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAuditEventDAO.java
```diff
@@ -61,6 +60,37 @@ protected AuditEventCriteriaBuilder entityKey(final String entityKey) {
...
+        protected static String toLikePattern(final String value) {
+            return value.
+                    replace("#", "##").
+                    replace("%", "#%").
+                    replace("_", "#_").
+                    replace("*", "%");
+        }
+
+        public AuditEventCriteriaBuilder who(final Set<String> who, final List<Object> parameters) {
+            if (who != null && !who.isEmpty()) {
```
ilgrosso: replace with Spring's `CollectionUtils#isEmpty`

core/persistence-jpa/src/main/java/org/apache/syncope/core/persistence/jpa/dao/JPAAuditEventDAO.java
```diff
@@ -61,6 +60,37 @@ protected AuditEventCriteriaBuilder entityKey(final String entityKey) {
...
+                query.append(andIfNeeded()).append("(");
+                boolean first = true;
+                for (String value : who) {
```
ilgrosso: Can't this loop be replaced by stream with map and `Collectors#joining`?

ilgrosso (changes requested):
This PR is proposing to add the capability to search audit events by `who`, e.g. by the administrator that was logged as the requester for the action that was logged as audit entry.

I believe this is a nice addition but it will require a different issue on JIRA.

This because [SYNCOPE-1978](https://issues.apache.org/jira/browse/SYNCOPE-1978) is instead asking to restore the possibility to match audit events by `username`, e.g. to look for users whose payload was logged for the audit entry - as opposite as the current capability to look for such users only by their `key`.

ozimakov:
Hi @ilgrosso
Thank you so much for the eye-opening comment. It's always pretty refreshing to see that you've been following the false goal, haha.
I've created a new ticket and rerouted this PR against it. Also fixed the issues you've outlined. Hopefully, it still might be useful. 
And I'll be working on the original issue (SYNCOPE-1978) in a separate PR.

ozimakov:
FYI: Looking ahead into SYNCOPE-1978, especially considering current approach on Elastic/OpenSearch term indexing, I decided to drop the requirement on the wildcard search here as well - to have similar API for "username" and for "who". Now, it is only exact match, including ability for multiple values (related via "OR"), is supported.

ilgrosso (approved):

ilgrosso:
Thank you @ozimakov I have cherry-picked this work to branches `4_1_X` and `4_0_X`.

PR #1441, opened 2026-06-26
Fix close button double icon

TatoniMatteo:
The "X" icon of close buttons was doubled.


ilgrosso (approved):

ilgrosso:
@TatoniMatteo is the same fix also needed by `4_1_X` and `4_0_X`?

PR #1515, opened 2026-09-01
[SYNCOPE-1993] provide support for enable auto registration parameter also in core module

andrea-patricelli:


core/starter/src/main/java/org/apache/syncope/core/starter/SyncopeCoreApplication.java
```diff
@@ -122,14 +123,16 @@ public TaskExecutorUnloader taskExecutorUnloader(final ListableBeanFactory beanF
...
+            final KeymasterProperties keymasterProps) {
+        return new SyncopeCoreStart(domainHolder, keymasterProps.isEnableAutoRegistration());
     }
 
     @ConditionalOnMissingBean
     @Bean
-    public KeymasterStop keymasterStop(final DomainHolder<?> domainHolder) {
-        return new SyncopeCoreStop(domainHolder);
+    public KeymasterStop keymasterStop(final DomainHolder<?> domainHolder, final KeymasterProperties keymasterProps) {
+        return new SyncopeCoreStop(domainHolder, keymasterProps);
```
ilgrosso: Why passing only `boolean` to `SyncopeCoreStart` constructor and the whole `KeymasterProperties` to `SyncopeCoreStop`?

ilgrosso (changes requested):

core/starter/src/main/java/org/apache/syncope/core/starter/SyncopeCoreApplication.java
andrea-patricelli: This is nonsense, going to align

ilgrosso (approved):

